PR #6, opened 2020-03-16
Add `Zip` for multiple Publishers

jasdev:
Rough sketch of overloads to allow folks to zip arbitrarily many publishers. I tried to mirror the existing style in the repository, but please feel free to send any and all feedback (no nit too small)! 🙏🏽

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ odd: 5
 
 `ZipMany.swift` contains two overloads on Combine’s `Publisher.zip` methods (which, at the time of writing only go up to arity three).
 
-This allows folks to zip arbitrarily many publishers and receive either an array of inner publisher outputs back.
+This lets you arbitrarily zip many publishers and receive either an array of inner publisher outputs back.
 
 ```swift
 let first = PassthroughSubject<Int, Never>()
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ All operators, utilities and helpers respect Combine's publisher contract, inclu
 * [values](#values)
 * [failures](#failures)
 * [dematerialize](#dematerialize)
-* [zip(with:) and zip(with:transform:)](#ZipMany)
+* [zip(with:)](#ZipMany)
 
 ### Publishers
 * [AnyPublisher.create](#anypublisher.create)
@@ -272,9 +272,9 @@ odd: 5
 
 ### ZipMany
 
-`ZipMany.swift` contains two pairs of overloads on Combine’s `Publisher.zip` methods (which, at the time of writing only go up to arity three).
+`ZipMany.swift` contains two overloads on Combine’s `Publisher.zip` methods (which, at the time of writing only go up to arity three).
 
-This allows folks to zip arbitrarily many publishers and receive either an array of inner publisher outputs—or transformed values—back.
+This allows folks to zip arbitrarily many publishers and receive either an array of inner publisher outputs back.
 
 ```swift
 let first = PassthroughSubject<Int, Never>()
@@ -283,7 +283,8 @@ let third = PassthroughSubject<Int, Never>()
 let fourth = PassthroughSubject<Int, Never>()
 
 subscription = first
-  .zip(with: second, third, fourth, transform: { $0.reduce(0, +) })
+  .zip(with: second, third, fourth)
+  .map { $0.reduce(0, +) }
   .sink(receiveValue: { print("zipped: \($0)") })
 
 first.send(1)
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ All operators, utilities and helpers respect Combine's publisher contract, inclu
 * [values](#values)
 * [failures](#failures)
 * [dematerialize](#dematerialize)
+* [zip(with:) and zip(with:transform:)](#ZipMany)
 
 ### Publishers
 * [AnyPublisher.create](#anypublisher.create)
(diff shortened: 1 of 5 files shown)


freak4pc:
image

Sources/Operators/ZipMany.swift
```diff
@@ -0,0 +1,54 @@
...
+        -> AnyPublisher<[Output], Failure> where Other.Output == Output, Other.Failure == Failure {
+        others.reduce(
+            map { [$0] }
+            .eraseToAnyPublisher()) { zipped, next in
+            zipped
+                .zip(next)
+                .map { $0.0 + [$0.1] }
+                .eraseToAnyPublisher()
+        }
+        .eraseToAnyPublisher()
```
freak4pc: Mind formatting like this instead? Was a tad hard to read

```suggestion
        let seed = map { [$0] }.eraseToAnyPublisher()

        return others
            .reduce(seed) { zipped, next in
                zipped
                    .zip(next)
                    .map { $0.0 + [$0.1] }
                    .eraseToAnyPublisher()
            }
            .eraseToAnyPublisher()
```

Sources/Operators/ZipMany.swift
```diff
@@ -0,0 +1,54 @@
...
+import Combine
+
+public extension Publisher {
+    // MARK: - Non-transform zipping variants.
+
+    /// Zips `self` with an array’s-worth of publishers with the same output and failure types.
+    /// Since there can be any number of `others`, `[Output]` values are emitted after zipping.
+    /// - Parameter others: The other publishers to zip with.
+    ///
+    /// - returns: A type-erased publisher with value events from each of the inner publishers zipped together in an array.
```
freak4pc:
```suggestion
    /// Zips `self` with an array of publishers with the same output and failure types.
    ///
    /// Since there can be any number of `others`, arrays of `Output` values are emitted after zipping.
    ///
    /// - parameter others: The other publishers to zip with.
    ///
    /// - returns: A type-erased publisher with value events from each of the inner publishers zipped together in an array.
```

Sources/Operators/ZipMany.swift
```diff
@@ -0,0 +1,54 @@
...
+            zip(with: others).map(transform)
+    }
+
+    /// A variadic overload on `Publisher.zip(with:transform)`.
+    func zip<Other: Publisher, Transformed>(with others: Other..., transform: @escaping ([Output]) -> Transformed)
+        -> Publishers.Map<AnyPublisher<[Output], Failure>, Transformed>
+        where Other.Output == Output, Other.Failure == Failure {
+            zip(with: others).map(transform)
+    }
+}
```
freak4pc: I'd personally opt for removing the transform overloads altogether, it's the same as just adding `.map` immediately after (which is what you do), and it would have more context if the consumer just do that themselves

```suggestion
```

Tests/ZipManyTests.swift
```diff
@@ -0,0 +1,131 @@
...
+        second.send(2)
+        third.send(3)
+
+        XCTAssertEqual(results, [[1, 2, 3]])
+        XCTAssertFalse(completed)
+        first.send(completion: .finished)
+        XCTAssertTrue(completed)
+    }
+
+    func testMultipleEmissionZippingEndingWithAnError() {
```
freak4pc: can we add one more tests one one of the publishers complete with `.finished` ?

Sources/Operators/ZipMany.swift
freak4pc: some minor formatting/word-tweaking, not sure why the diff makes it seem like I changed the whole thing :]

Tests/ZipManyTests.swift
jasdev: Sure thing! I’ll add it alongside a couple of other tidying-up commits. 👌🏽

jasdev:
> I’ve had a couple of rough days with the entire COVID crisis, sorry this took a bit longer than expected.

No worries at all! Hope any dust has settled and that y’all are safe. Never any rush to reply—it’s already inspiring that you’re able to find the time, even as a new parent.

I think I covered everything (sorry for the force push, had a rebasing snag)! Let me know if there’s anything else blocking this PR, when you get some headspace. 🙏🏽

freak4pc:
I'm a fan of force-push on feature branches, no need to apologize :) 
Thank you!

Tests/ZipManyTests.swift
```diff
@@ -0,0 +1,149 @@
...
+
+        second.send(2)
+        second.send(2)
+
+        third.send(3)
+
+        XCTAssertEqual(results, [[1, 2, 3]])
+        XCTAssertNil(completed)
+        first.send(completion: .finished) // Triggers a completion, since, there
+        // aren’t any buffered events from `first` (or `third`) to possible pair with.
```
freak4pc:
```suggestion
        // aren’t any buffered events from `first` (or `third`) to possibly pair with.
```

README.md
```diff
@@ -267,6 +268,37 @@ even: 4
 odd: 5
 ```
 
+------
+
+### ZipMany
+
+`ZipMany.swift` contains two overloads on Combine’s `Publisher.zip` methods (which, at the time of writing only go up to arity three).
+
+This allows folks to zip arbitrarily many publishers and receive either an array of inner publisher outputs back.
```
freak4pc:
```suggestion
This lets you arbitrarily zip many publishers and receive either an array of inner publisher outputs back.
```

README.md
```diff
@@ -267,6 +268,37 @@ even: 4
 odd: 5
 ```
 
+------
+
+### ZipMany
+
+`ZipMany.swift` contains two overloads on Combine’s `Publisher.zip` methods (which, at the time of writing only go up to arity three).
```
freak4pc: Uhm sorryp 
```suggestion
This repo includes two overloads on Combine’s `Publisher.zip` methods (which, at the time of writing only go up to arity three).
```

PR #7, opened 2020-03-22
First pass at `Publisher.removeAllDuplicates` and `removeAllDuplicates(by:)`.

jasdev:
Tidying up and PR’ing `Publisher.removeAllDuplicates` and `.removeAllDuplicates(by:)` from an earlier sketch [on Twitter](https://twitter.com/jasdev/status/1214764966974312448) and as per your mention of [`Observable.distinct` and `.distinct(_:)`](https://github.com/RxSwiftCommunity/RxSwiftExt/blob/1e7a8683f3a8960042caa9b0ad0fc41f1217104d/Source/RxSwift/distinct.swift#L12-L69).

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,10 +28,12 @@ All operators, utilities and helpers respect Combine's publisher contract, inclu
 * [values](#values)
 * [failures](#failures)
 * [dematerialize](#dematerialize)
+* [partition](#partition)
 * [zip(with:) and Collection.zip](#ZipMany)
 * [combineLatest(with:) and Collection.combineLatest](#CombineLatestMany)
 * [mapMany(_:)](#MapMany)
 * [setOutputType(to:)](#setOutputType)
+* [removeAllDuplicates and removeAllDuplicates(by:) ](#removeAllDuplicates)
 
 ### Publishers
 * [AnyPublisher.create](#AnypublisherCreate)
@@ -312,7 +314,7 @@ odd: 5
 
 This repo includes two overloads on Combine’s `Publisher.zip` methods (which, at the time of writing only go up to arity three).
 
-This lets you arbitrarily zip many publishers and receive either an array of inner publisher outputs back.
+This lets you arbitrarily zip many publishers and receive an array of inner publisher outputs back.
 
 ```swift
 let first = PassthroughSubject<Int, Never>()
@@ -377,7 +379,6 @@ combineLatest: [true, true, true, true]
 combineLatest: [false, true, true, true]
 ```
 
-
 ### MapMany
 
 Projects each element of a publisher collection into a new publisher collection form.
@@ -401,6 +402,25 @@ intArrayPublisher.send([10, 2, 2, 4, 3, 8])
 ### setOutputType
 
 `Publisher.setOutputType(to:)` is an analog to [`.setFailureType(to:)`](https://developer.apple.com/documentation/combine/publisher/3204753-setfailuretype) for when `Output` is constrained to `Never`. This is especially helpful when chaining operators after an [`.ignoreOutput()`](https://developer.apple.com/documentation/combine/publisher/3204714-ignoreoutput) call.
+
+### removeAllDuplicates
+
+`Publisher.removeAllDuplicates` and `.removeAllDuplicates(by:)` are stricter forms of Apple’s [`Publisher.removeDuplicates`](https://developer.apple.com/documentation/combine/publisher/3204745-removeduplicates) and [`.removeDuplicates(by:)`](https://developer.apple.com/documentation/combine/publisher/3204746-removeduplicates)—the operators de-duplicate across _all_ previous value events, instead of pairwise.
+
+If your `Output` doesn‘t conform to `Hashable` or `Equatable`, you may instead use the predicate-based version of this operator to decide whether two elements are equal.
+
+```swift
+subscription = [1, 1, 2, 1, 3, 3, 4].publisher
+  .removeAllDuplicates()
+  .sink(receiveValue: { print("removeAllDuplicates: \($0)") })
+```
+
+```none
+removeAllDuplicates: 1
+removeAllDuplicates: 2
+removeAllDuplicates: 3
+removeAllDuplicates: 4
+```
 
 ## Publishers
 
--- a/Sources/Operators/RemoveAllDuplicates.swift
+++ b/Sources/Operators/RemoveAllDuplicates.swift
@@ -1,0 +1,63 @@
+//
+//  RemoveAllDuplicates.swift
+//  CombineExt
+//
+//  Created by Jasdev Singh on 21/03/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import Combine
+
+public extension Publisher where Output: Hashable {
+    /// De-duplicates _all_ published value events, as opposed
+    /// to pairwise with `Publisher.removeDuplicates`.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values
+    ///         in an in-memory `Set`. So, use this operator with caution, when handling publishers
+    ///         that emit a large number of unique value events.
+    ///
+    /// - returns: A publisher that consumes duplicate values across all previous emissions from upstream.
+    func removeAllDuplicates() -> Publishers.Filter<Self> {
+        var seen = Set<Output>()
+        return filter { incoming in seen.insert(incoming).inserted }
+    }
+}
+
+public extension Publisher where Output: Equatable {
+    /// `Publisher.removeAllDuplicates` de-duplicates _all_ published `Hashable`-conforming value events, as opposed to pairwise with `Publisher.removeDuplicates`.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values in an in-memory `Array`. So, use
+    ///         this operator with caution, when handling publishers that emit a large number of unique value events.
+    ///
+    /// - returns: A publisher that consumes duplicate values across all previous emissions from upstream.
+    func removeAllDuplicates() -> Publishers.Filter<Self> {
+        removeAllDuplicates(by: ==)
+    }
+}
+
+public extension Publisher {
+    /// De-duplicates _all_ published value events, along the provided `by` predicate, as opposed to pairwise with `Publisher.removeDuplicates(by:)`.
+    ///
+    /// - parameter by: A predicate to use when determining uniqueness. `Publisher.removeAllDuplicates` will iterate
+    ///                 over all seen values applying each known unique value as the first argument to the predicate and the
+    ///                 incoming value event as the second, i.e. `by(see, next) -> Bool`. If this predicate is `true` for any
+    ///                 seen value, the next incoming value isn’t emitted downstream.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values
+    ///         in an in-memory `Array`. So, use this operator with caution, when handling publishers
+    ///         that emit a large number of unique value events (as per `by`).
+    ///
+    /// - returns: A publisher that consumes duplicate values across all previous emissions from upstream
+    ///            (signaled with `by`).
+    func removeAllDuplicates(by comparator: @escaping (Output, Output) -> Bool) -> Publishers.Filter<Self> {
+        var seen = [Output]()
+        return filter { incoming in
+            if seen.contains(where: { comparator($0, incoming) }) {
+                return false
+            } else {
+                seen.append(incoming)
+                return true
+            }
+        }
+    }
+}
--- a/CombineExt.xcodeproj/project.pbxproj
+++ b/CombineExt.xcodeproj/project.pbxproj
@@ -47,6 +47,8 @@
 		AAEAF0E92436D785007C35E0 /* SetOutputTypeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AAEAF0E82436D785007C35E0 /* SetOutputTypeTests.swift */; };
 		BF50924B241FFE8E00600DF4 /* ZipManyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF50924A241FFE8E00600DF4 /* ZipManyTests.swift */; };
 		BF8121BC241FF42C006A93B8 /* ZipMany.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF8121BB241FF42C006A93B8 /* ZipMany.swift */; };
+		BF84B7412426B786001BFA88 /* RemoveAllDuplicates.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF84B7402426B786001BFA88 /* RemoveAllDuplicates.swift */; };
+		BF84B7432426C332001BFA88 /* RemoveAllDuplicatesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF84B7422426C332001BFA88 /* RemoveAllDuplicatesTests.swift */; };
 		BFB4EA132428256B0096E9E9 /* CombineLatestMany.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFB4EA122428256B0096E9E9 /* CombineLatestMany.swift */; };
 		BFB4EA1524283ECF0096E9E9 /* CombineLatestManyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFB4EA1424283ECF0096E9E9 /* CombineLatestManyTests.swift */; };
 		DC16910F24281A1800B234C4 /* MapMany.swift in Sources */ = {isa = PBXBuildFile; fileRef = DC16910E24281A1800B234C4 /* MapMany.swift */; };
@@ -93,6 +95,8 @@
 		AAEAF0E82436D785007C35E0 /* SetOutputTypeTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SetOutputTypeTests.swift; sourceTree = "<group>"; };
 		BF50924A241FFE8E00600DF4 /* ZipManyTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ZipManyTests.swift; sourceTree = "<group>"; };
 		BF8121BB241FF42C006A93B8 /* ZipMany.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ZipMany.swift; sourceTree = "<group>"; };
+		BF84B7402426B786001BFA88 /* RemoveAllDuplicates.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RemoveAllDuplicates.swift; sourceTree = "<group>"; };
+		BF84B7422426C332001BFA88 /* RemoveAllDuplicatesTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RemoveAllDuplicatesTests.swift; sourceTree = "<group>"; };
 		BFB4EA122428256B0096E9E9 /* CombineLatestMany.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CombineLatestMany.swift; sourceTree = "<group>"; };
 		BFB4EA1424283ECF0096E9E9 /* CombineLatestManyTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CombineLatestManyTests.swift; sourceTree = "<group>"; };
 		"CombineExt::CombineExt::Product" /* CombineExt.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; path = CombineExt.framework; sourceTree = BUILT_PRODUCTS_DIR; };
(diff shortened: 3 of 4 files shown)


jasdev:
(weird, code coverage is showing the `Equatable` variant isn’t be covered—investigating 🔍 )

jasdev:
[> (An enum without associated values has Hashable conformance even without the declaration.)](https://developer.apple.com/documentation/swift/hashable)

Whelp, TIL!

Sources/Operators/RemoveKnownDuplicates.swift
```diff
@@ -0,0 +1,76 @@
...
+    /// published value events, as opposed to pairwise with `Publisher.removeDuplicates`.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values
+    ///         in an in-memory `Set`. So, use this operator with caution, when handling publishers
+    ///         that emit a large number of unique value events.
+    ///
+    /// - returns: A type-erased publisher that consumes duplicate values across all previous emissions from upstream.
+    func removeKnownDuplicates() -> AnyPublisher<Output, Failure> {
+        scan((Set<Output>(), Output?.none)) { seenAndPrevious, incoming in
+            var seen = seenAndPrevious.0
```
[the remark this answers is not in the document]
jasdev: Best GIF (cc @krausefx)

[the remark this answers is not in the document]
freak4pc: this is what i do when I can't find time to review
find a meme-able line of code, put a GIF, move on with my day 
💰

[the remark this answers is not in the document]
jasdev: Best move 😂

Sources/Operators/RemoveKnownDuplicates.swift
```diff
@@ -0,0 +1,76 @@
...
+//  CombineExt
+//
+//  Created by Jasdev Singh on 21/03/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import Combine
+
+public extension Publisher where Output: Hashable {
+    /// When `Output` conforms to `Hashable`, `Publisher.removeKnownDuplicates` de-duplicates _all_
```
freak4pc:
```suggestion
    /// De-duplicates _all_
```

The generic constraint won't let you use this otherwise so I'd remove it from here and let folks immediately get to the beef of "what does this do"

README.md
```diff
@@ -308,6 +310,27 @@ You may also use `.zip()` directly on an array of publishers with the same outpu
 zipped: 10
 ```
 
+### removeKnownDuplicates
+
+`Publisher.removeKnownDuplicates` and `.removeKnownDuplicates(by:)` are stricter forms of Apple’s [`Publisher.removeDuplicates`](https://developer.apple.com/documentation/combine/publisher/3204745-removeduplicates) and [`.removeDuplicates(by:)`](https://developer.apple.com/documentation/combine/publisher/3204746-removeduplicates)—the operators de-duplicate across _all_ previous value events, instead of pairwise.
+
```
freak4pc:
```suggestion
```

remove extra line

README.md
```diff
@@ -308,6 +310,27 @@ You may also use `.zip()` directly on an array of publishers with the same outpu
...
+
+There are overloads for when `Output` either conforms to `Hashable` or `Equatable` and a predicate-based, `(Output) -> Bool` version is also provided.
```
freak4pc: WDYT of :
```suggestion
If your `Output` doesn't conform to `Hashable` or `Equatable`, you may instead use the predicate-based version of this operator to decide whether two elements are indeed equal.
```

Sources/Operators/RemoveKnownDuplicates.swift
```diff
@@ -0,0 +1,76 @@
...
+            return (seen, isUnique ? incoming : nil)
+        }
+        .compactMap { $0.1 }
+        .eraseToAnyPublisher()
+    }
+}
+
+public extension Publisher where Output: Equatable {
+    /// When `Output` conforms to `Equatable`, `Publisher.removeKnownDuplicates` de-duplicates _all_
+    /// published value events, as opposed to pairwise with `Publisher.removeDuplicates`.
```
freak4pc: Same re-docs. If we're noting specifically re-Hashable/Equatable, it's better to put it at the end of the sentence, like 

(just with better line break lol)
```suggestion
    /// `Publisher.removeKnownDuplicates` de-duplicates _all_
    /// published `Hashable`-conforming value events, as opposed to pairwise with `Publisher.removeDuplicates`.
```

Sources/Operators/RemoveKnownDuplicates.swift
```diff
@@ -0,0 +1,76 @@
...
+    /// `Publisher.removeKnownDuplicates(by:)` de-duplicates _all_ published value events,
+    /// along the provided `by` predicate, as opposed to pairwise with `Publisher.removeDuplicates(by:)`.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values
+    ///         in an in-memory `Array`. So, use this operator with caution, when handling publishers
+    ///         that emit a large number of unique value events (as per `by`).
+    ///
+    /// - returns: A type-erased publisher that consumes duplicate values across all previous emissions from upstream
+    ///            (signaled with `by`).
+    func removeKnownDuplicates(by predicate: @escaping (Output) -> Bool) -> AnyPublisher<Output, Failure> {
```
freak4pc:
```suggestion
    func removeKnownDuplicates(by predicate: @escaping (Output) -> Bool) -> AnyPublisher<Output, Failure> {
```

I'm wondering here, is it not possible to implement the Equatable and Hashable ones by simply using this one and implementing the predicate? I know it's not like that in Ext but wondering if it's possible to save some _duplicate_ [pun-intended] implementation 🤔

Tests/RemoveKnownDuplicatesTests.swift
```diff
@@ -0,0 +1,224 @@
...
+        case anError
+    }
+
+    // MARK: - `Hashable`-related tests
+
+    func testHashableExpectedDeduplication() {
+        var results = [Int]()
+
+        subscription = [1, 1, 2, 1, 3, 3, 4].publisher
+            .removeKnownDuplicates()
```
freak4pc: Maybe create `Hashable`-conforming and `Equatable`-conforming structs separately here?

Sources/Operators/RemoveKnownDuplicates.swift
```diff
@@ -0,0 +1,76 @@
...
+//  RemoveKnownDuplicates.swift
+//  CombineExt
+//
+//  Created by Jasdev Singh on 21/03/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import Combine
+
+public extension Publisher where Output: Hashable {
```
freak4pc: I'm wondering if this is eqtually needed, given

```swift
public protocol Hashable : Equatable
```

Wouldn't the Equatable conformance basically deal with `Hashable` as well?

freak4pc:
P.S. I'm feeling a bit iffy about the name, but don't have a better idea yet :D

jasdev:
> P.S. I’m feeling a bit iffy about the name, but don’t have a better idea yet :D
perhaps `removeDistinctDuplicates`?

Kinda leaning `distinct` (à la Rx’s naming), in that case 😄

Sources/Operators/RemoveKnownDuplicates.swift
jasdev: oo great call, let me give it a try!

jasdev: It worked! A note though, I had to refactor the `by` predicate to follow [Combine’s lead](https://developer.apple.com/documentation/combine/publisher/3204746-removeduplicates) since the only way to lean on `==` is to make the predicate two-ary.

Sources/Operators/RemoveKnownDuplicates.swift
jasdev: It could, but I think it might be worth the two separate implementations. The `Output: Hashable` constraint lets us `scan` with a `Set` instead of an `Array`—so we get a slight speed bump to constant time membership checks instead of linear.

If you don’t think the parallel implementations are worth the performance gain, also open to removing it! 💯

jasdev:
(Should be good for another pass over! Thanks for the super quick turnaround on this. 🙇🏽‍♂️)

Sources/Operators/Distinct.swift
```diff
@@ -0,0 +1,72 @@
...
+    /// `Publisher.distinct` de-duplicates _all_
+    /// published `Hashable`-conforming value events, as opposed to pairwise with `Publisher.removeDuplicates`.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values
+    ///         in an in-memory `Array`. So, use this operator with caution, when handling publishers
+    ///         that emit a large number of unique value events.
+    ///
+    /// - returns: A type-erased publisher that consumes duplicate values across all previous emissions from upstream.
+    func distinct() -> AnyPublisher<Output, Failure> {
+        distinct(by: ==)
```
jasdev: :nodg:

Sources/Operators/RemoveKnownDuplicates.swift
freak4pc: That's ok :)

Sources/Operators/RemoveKnownDuplicates.swift
freak4pc: Gotya, that's OK :) Let's leave as is.

freak4pc:
> Kinda leaning distinct (à la Rx’s naming),

I don't think the name fits here... In RxSwift you have distinctUntilChanged, so distinct makes sense. I think it would be beneificial to lean on the Combine naming here. Any other thoughs?

jasdev:
> Any other thoughs?

Hmm,

- `Publisher.removeRollingDuplicates`
- `.removeAllDuplicates`
- `.removeSeenDuplicates`

(thinking aloud)—the only reason I hesitate with `.removeDistinctDuplicates` is that, at glance, “distinct” and “duplicates” are antonyms that sit next to one other in the name. :thinking-mergesort:

freak4pc:
Another weird thought :) 

`removeDuplicates(distinct: Bool)` 
   true -> use your implementations
   false -> relay to regular `removeDuplicates`

jasdev:
We could! The only downside I can see from that approach is the `false` path would require us to type-erase `Publishers.RemoveDuplicates` and I’m not sure if Apple has any specific, fused operator implementations for it that we’d be cutting off. 🤔

No strong opinions here—I’m game to roll with whatever name you like best (“best name”).

Sources/Operators/Distinct.swift
```diff
@@ -0,0 +1,72 @@
...
+    ///         in an in-memory `Array`. So, use this operator with caution, when handling publishers
+    ///         that emit a large number of unique value events (as per `by`).
+    ///
+    /// - returns: A type-erased publisher that consumes duplicate values across all previous emissions from upstream
+    ///            (signaled with `by`).
+    func distinct(by predicate: @escaping (Output, Output) -> Bool) -> AnyPublisher<Output, Failure> {
+        scan(([Output](), Output?.none)) { seenAndPrevious, incoming in
+            var seen = seenAndPrevious.0
+
+            guard !seen.contains(where: { predicate($0, incoming) }) else { return (seen, nil) }
```
jdisho: nit: Personally, I'm not a fan of guarding against a negation, I find it a bit harder to reason. 
I think `if` instead  would make this a bit more simple 😊

jasdev: Agreed, updating this in the next commit. Thanks. 🙏🏽

jasdev:
@jdisho,

> Why are we type-erasing some publisher and not some others? Is there a motivation behind?

Great call—this specific PR doesn’t need type erasure (and I can fix that in the next commit). Previous operators were forced to, since conditionals caused different `Publisher` conformances to be returned in certain cases, e.g. [the `switch` in `Collection.zip`](https://github.com/CombineCommunity/CombineExt/blob/a5729f3dff0fd7c8572593decba4e92a26f39e5b/Sources/Operators/ZipMany.swift#L50-L60). The way to get around that in those scenarios (I think) is to fully roll your own non-composed operator, which is a lot to take on and fusion only kicks in, if folks then add overloads for specific operators on that newly created type.

freak4pc:
@jdisho I actually think that we _should_ type-erase composed operators. There is no reason for a consumer to know we used 3 operators to make this one, for example. [they really don't care] 

This is more about ergonomics and aesthetics than it is anything "technical". 

It would've been really nice having a typed operator for this kind of thing, but this requires a custom operator implementation, like in `Materialize` or `WithLatestFrom`, in these cases you just return `Publishers.Materialize` or `Publishers.WithLatestFrom`. This isn't super hard with the `DemandBuffer` and `Sink` this repo provides, and still, I understand the simplicity of existing operator composition. 

I'd say that we can make an unwritten rule that if a custom operator uses _other_ operators (especially more than one), we should erase the output to hide the non-interesting implementation details. 


Thoughts @jdisho @jasdev ?

jdisho:
Makes sense to me @freak4pc. 👍

jasdev:
> Makes sense to me @freak4pc.

+1! hmu with any thoughts on the final name here and I’ll update the diff., so we can land this when @freak4pc is ready. 🙏🏽

jdisho:
> +1! hmu with any thoughts on the final name here and I’ll update the diff., so we can land this when @freak4pc is ready. 🙏🏽

Awesome. Thank you 🙏

jdisho:
> > +1! hmu with any thoughts on the final name here and I’ll update the diff., so we can land this when @freak4pc is ready. 🙏🏽
> 
> Awesome! Thank you 🙏

jasdev:
(No rush @freak4pc—thiiink I covered all the blockers, so it should be good for another review when you have some time. 🙇🏽‍♂️)

freak4pc:
I'll take a peek at this tomorrow, hopefully! Thanks :)

jasdev:
(hmmm—we might need to bump CombineExt’s stack, now that 11.4 is out :nodg:)

jasdev:
Related: https://github.com/actions/virtual-environments/issues/620

jasdev:
(reverting that `compactMap` key path change for now to keep this passing CI.)

Sources/Operators/Distinct.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// - returns: A publisher that consumes duplicate values across all previous emissions from upstream.
+    func distinct() -> Publishers.Filter<Self> {
+        var seen = Set<Output>()
+        return filter { incoming in seen.insert(incoming).inserted }
+    }
+}
+
+public extension Publisher where Output: Equatable {
+    /// `Publisher.distinct` de-duplicates _all_
+    /// published `Hashable`-conforming value events, as opposed to pairwise with `Publisher.removeDuplicates`.
```
freak4pc, line 28: The line breaks on this first line are a tad weird, like the other ones

Sources/Operators/Distinct.swift
```diff
@@ -0,0 +1,66 @@
...
+    ///
+    /// - returns: A publisher that consumes duplicate values across all previous emissions from upstream.
+    func distinct() -> Publishers.Filter<Self> {
+        distinct(by: ==)
+    }
+}
+
+public extension Publisher {
+    /// `Publisher.distinct(by:)` de-duplicates _all_ published value events,
+    /// along the provided `by` predicate, as opposed to pairwise with `Publisher.removeDuplicates(by:)`.
```
freak4pc, line 42:
```suggestion
    /// `Publisher.distinct(by:)` de-duplicates _all_ published value events, where their equality is determined by the provided closure
```

Sources/Operators/Distinct.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// incoming value event as the second, i.e. `by(see, next) -> Bool`. If this predicate is `true` for any
+    /// seen value, the next incoming value isn’t emitted downstream.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values
+    ///         in an in-memory `Array`. So, use this operator with caution, when handling publishers
+    ///         that emit a large number of unique value events (as per `by`).
+    ///
+    /// - returns: A publisher that consumes duplicate values across all previous emissions from upstream
+    ///            (signaled with `by`).
+    func distinct(by predicate: @escaping (Output, Output) -> Bool) -> Publishers.Filter<Self> {
```
freak4pc, line 55:
```suggestion
    func distinct(by comparator: @escaping (Output, Output) -> Bool) -> Publishers.Filter<Self> {
```

I think a predicate is when you're filtering for something, but here you're comparing elements

README.md
```diff
@@ -362,6 +363,25 @@ intArrayPublisher.send([10, 2, 2, 4, 3, 8])
 ["10", "2", "2", "4", "3", "8"]
 ```
 
+### Distinct
+
+`Publisher.distinct` and `.distinct(by:)` are stricter forms of Apple’s [`Publisher.removeDuplicates`](https://developer.apple.com/documentation/combine/publisher/3204745-removeduplicates) and [`.removeDuplicates(by:)`](https://developer.apple.com/documentation/combine/publisher/3204746-removeduplicates)—the operators de-duplicate across _all_ previous value events, instead of pairwise.
```
freak4pc, line 368: The biggest issue I have is still with the name - `distinct` operators are from Rx, where here we might want to keep the same naming convention. My currrent instinct is something like `removeUniqueDuplicates`, `removeAllDuplicates` or `removeDuplicatesUniquely` but I'm very open to other variations, preferably coming from the same family.

Sources/Operators/Distinct.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// De-duplicates _all_
+    /// published value events, as opposed to pairwise with `Publisher.removeDuplicates`.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values
+    ///         in an in-memory `Set`. So, use this operator with caution, when handling publishers
+    ///         that emit a large number of unique value events.
+    ///
+    /// - returns: A publisher that consumes duplicate values across all previous emissions from upstream.
+    func distinct() -> Publishers.Filter<Self> {
+        var seen = Set<Output>()
```
freak4pc, line 21: There's something I hadn't considered. A set determines uniqueness by Hasability. That's why the definition of it is

`public struct Set<Element> where Element: Hashable {`

Wouldn't this basically mean you can't use this operator for anything that isn't Hashable-cconforming?

Sources/Operators/Distinct.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// `Publisher.distinct(by:)` de-duplicates _all_ published value events,
+    /// along the provided `by` predicate, as opposed to pairwise with `Publisher.removeDuplicates(by:)`.
+    ///
+    /// - parameter by: A predicate to use when determining uniqueness. `Publisher.distinct` will iterate
+    /// over all seen values applying each known unique value as the first argument to the predicate and the
+    /// incoming value event as the second, i.e. `by(see, next) -> Bool`. If this predicate is `true` for any
+    /// seen value, the next incoming value isn’t emitted downstream.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values
+    ///         in an in-memory `Array`. So, use this operator with caution, when handling publishers
```
freak4pc, line 50:
```suggestion
    ///         in an in-memory `Set`. So, use this operator with caution, when handling publishers
```

README.md
jasdev: Game for `removeAllDuplicates`, I’ll rename everything `distinct`-related in an upcoming commit. 😎

Sources/Operators/Distinct.swift
jasdev: That’s where the more relaxed `Equatable` and comparator-based overloads come in. :nodg:

Sources/Operators/Distinct.swift
```diff
@@ -0,0 +1,66 @@
...
+        return filter { incoming in seen.insert(incoming).inserted }
+    }
+}
+
+public extension Publisher where Output: Equatable {
+    /// `Publisher.distinct` de-duplicates _all_
+    /// published `Hashable`-conforming value events, as opposed to pairwise with `Publisher.removeDuplicates`.
+    ///
+    /// - note: It’s important to note that this operator stores all emitted values
+    ///         in an in-memory `Array`. So, use this operator with caution, when handling publishers
```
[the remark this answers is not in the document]
jasdev: It’s subtle, but the original `Array` mention here is right (and below). Once we relax the `Hashable` constraint, the best we can do is use an in-memory `Array`.

Sources/Operators/Distinct.swift
jasdev: (Same as above. 😌)

jasdev:
@freak4pc: Thiiink I got it all—definitely feel free to let me know if the doc. line breaks are wonky. 👌🏽 Tried my best to balance line length and readability.

Sources/Operators/RemoveAllDuplicates.swift
```diff
@@ -0,0 +1,63 @@
...
+    ///         this operator with caution, when handling publishers that emit a large number of unique value events.
+    ///
+    /// - returns: A publisher that consumes duplicate values across all previous emissions from upstream.
+    func removeAllDuplicates() -> Publishers.Filter<Self> {
+        removeAllDuplicates(by: ==)
+    }
+}
+
+public extension Publisher {
+    /// `Publisher.removeAllDuplicates(by:)` de-duplicates _all_ published value events, along the provided `by` predicate, as opposed to pairwise with `Publisher.removeDuplicates(by:)`.
```
freak4pc, line 39:
```suggestion
    /// De-duplicates _all_ published value events, along the provided `by` predicate, as opposed to pairwise with `Publisher.removeDuplicates(by:)`.
```

PR #8, opened 2020-03-22
Add RetryWhen

danielt1263:
Added RetryWhen with a couple of tests. And thank you for this repo which helped me understand how to create Publishers.

### DIFF
--- a/CombineExt.xcodeproj/project.pbxproj
+++ b/CombineExt.xcodeproj/project.pbxproj
@@ -21,6 +21,8 @@
 /* End PBXAggregateTarget section */
 
 /* Begin PBXBuildFile section */
+		72FB75602426F93400BE8D78 /* RetryWhen.swift in Sources */ = {isa = PBXBuildFile; fileRef = 72FB755F2426F93400BE8D78 /* RetryWhen.swift */; };
+		72FB75622426F95200BE8D78 /* RetryWhenTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 72FB75612426F95200BE8D78 /* RetryWhenTests.swift */; };
 		78002BB5241E910C0018AA28 /* Relay.swift in Sources */ = {isa = PBXBuildFile; fileRef = 78002BB4241E910C0018AA28 /* Relay.swift */; };
 		78002BB7241E915E0018AA28 /* CurrentValueRelay.swift in Sources */ = {isa = PBXBuildFile; fileRef = 78002BB6241E915E0018AA28 /* CurrentValueRelay.swift */; };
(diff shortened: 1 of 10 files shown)


danielt1263:
The real genius of this library are the `Sink` and `DemandBuffer` types. They make developing operators for Combine so much easier than it otherwise would be. However, the Sink type took an Upstream parameter for the sole purpose of performing work on it, work that IMO should not be done in an init method and work that stopped me from being able to reuse Sink from the retryWhen operator. I took the initiative and possibly too drastic step of removing that parameter. I also made retryWhen's parameter a bit more flexible. If I went too far with this change, I can back it out, but I really think it's for the better.

freak4pc:
Thanks for the kind words :) I'll check out you branch locally and try it a bit later, since this is a very important operator and also a big change to Sink.

My initial instinct is to not remove `upstream` since it serves >90% of use cases, but it _can_ potentially just be optional, so if it doesn't exist all of its dependencies won't occur. Don't make this change yet, I don't want you to work for nothing, I'll take a look later today or tomorrow hopefully, depending on what baby allows :D

danielt1263:
Another option would be to move the `upstream.subscribe(self)` into a `start()` method. I just need to be able to create the Sink _without_ starting it immediately.

danielt1263:
I have a concern about the Trigger of the retryWhen operator but I'm not sure how to fix it. As it stands, if the Trigger's Error type doesn't match the source's, then the error won't get reported. What I would like is for the Trigger's error type to be either Never or a type that matches the source and anything else would report a compile-time error. I'm not sure how to do that.

bok-:
Is there any appetite for this to be completed? I have a need for this in a project, is there anything I can do to help get it over the line?

freak4pc:
There is appetite but time is lacking with COVID making work/life/OSS balance a bit harder for many of us :)

mRs-:
I would like to have this merged as well, how can I help? Need to retryWhen now and Sink is not public or open 🙁

freak4pc:
Hey @danielt1263 - I'm sorry apparently this was auto-closed because I changed the primary branch to `main` from `master`, so everything was closed.

Would appreciate if you could re-open it rebased on top of `main`. Could you explain a bit on the motivation of removing `upstream` from Sink? It seems a bit counterintuitive, as usually the sink wants to be a "Man-in-the-middle" to navigate upstream signals to the downstream consumers.

danielt1263:
New pull request added. https://github.com/CombineCommunity/CombineExt/pull/101

I didn't have to make any modifications to Sink this time in order to re-add the operator.

PR #9, opened 2020-03-22
Add `mapMany(_:)` operator.

jdisho:
Basically doing the same thing as in the `RxSwiftExt` version of [`mapMany`](https://github.com/RxSwiftCommunity/RxSwiftExt/blob/master/Source/RxSwift/mapMany.swift). 

✌️

(Should update the `README`)

### DIFF
--- a/Sources/Operators/MapMany.swift
+++ b/Sources/Operators/MapMany.swift
@@ -1,0 +1,39 @@
+//
+//  MapMany.swift
+//  CombineExt
+//
+//  Created by Joan Disho on 22/03/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import Combine
+
+public extension Publisher where Output: Collection {
+    /// Projects each element of a publisher collection into a new publisher collection form.
+    ///
+    /// - parameter transform: A transformation function which applies to each element of the source collection.
+    ///
+    /// - returns: A publisher collection whose elements are the result of invoking the transformation function on each element of the source.
+    ///
+    /// An example usage could look as follows:
+        ///
+        ///    ```
+        ///    let intArrayPublisher = PassthroughSubject<[Int], Never>()
+        ///
+        ///    intArrayPublisher
+        ///         .mapMany(String.init)
+        ///         .sink(receiveValue: { print($0) })
+        ///
+        ///    intArrayPublisher.send([10, 2, 2, 4, 3, 8])
+        ///
+        ///    // Output: ["10", "2", "2", "4", "3", "8"]
+        ///    ```
+        ///
+    ///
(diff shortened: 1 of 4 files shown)


Sources/Operators/MapMany.swift
```diff
@@ -0,0 +1,36 @@
...
+
+public extension Publisher where Output: Collection {
+    /// Projects each element of a publisher collection into a new publisher collection form.
+    ///
+    /// - parameter transform: A transformation function which applies to each element of the source collection.
+    ///
+    /// - returns: A publisher collection whose elements are the result of invoking the transformation function on each element of the source.
+    ///
+    /// An example usage could look as follows:
+        ///
```
freak4pc: Mind aligning the `///` to the same horizontal placed? Looks a tad weird.

Sources/Operators/MapMany.swift
```diff
@@ -0,0 +1,36 @@
...
+        ///         .mapMany(String.init)
+        ///         .sink(receiveValue: { print($0) })
+        ///
+        ///    intArrayPublisher.send([10, 2, 2, 4, 3, 8])
+        ///
+        ///    // Output: ["10", "2", "2", "4", "3", "8"]
+        ///    ```
+        ///
+    ///
+    func mapMany<T>(_ transform: @escaping (Output.Element) -> T) -> Publishers.Map<Self, [T]> {
```
freak4pc:
```suggestion
    func mapMany<Result>(_ transform: @escaping (Output.Element) -> Result) -> Publishers.Map<Self, [Result]> {
```

Not a huge deal but maybe a bit nicer

freak4pc:
By the way do you think we should also have `tryMapMany`? not sure if that's too interesting

jdisho:
Updated the PR and addressed your points. @freak4pc  

Regarding `tryMapMany`, why not in an other PR?

freak4pc:
Well, because this one adds `mapMany` to mirror `map`, so we should do the same for the `try` one in that case (At least IMO...). Let's leave as is for now ;)

Sources/Operators/MapMany.swift
```diff
@@ -0,0 +1,36 @@
...
+    ///         .sink(receiveValue: { print($0) })
+    ///
+    ///    intArrayPublisher.send([10, 2, 2, 4, 3, 8])
+    ///
+    ///    // Output: ["10", "2", "2", "4", "3", "8"]
+    ///    ```
+    ///
+    ///
+    func mapMany<Result>(_ transform: @escaping (Output.Element) -> Result) -> Publishers.Map<Self, [Result]> {
+        return map { $0.map(transform) }
```
freak4pc:
```suggestion
        map { $0.map(transform) }
```

PR #12, opened 2020-03-27
Implementation of asFuture() operator

jullianm:
Hi!

I've implemented an operator that converts any publisher to a custom Future type that behaves like the Combine built-in one.

It produces a single value then finishes, or fails.

I thought this was interesting to create the RxSwift's asSingle() counterpart.

A sequence containing more than one element doesn't error out since it always asks for a single value through `subscription.request(.max(1))`

An empty sequence terminates with a `fatalError`.

Let me know your thoughts, thanks for reviewing!

Jullian

### DIFF
--- a/Sources/Operators/AsFuture.swift
+++ b/Sources/Operators/AsFuture.swift
@@ -1,0 +1,88 @@
+//
+//  AsFuture.swift
+//  CombineExt
+//
+//  Created by Jullianm on 27/03/2020.
+//  Copyright © 2020 Jullianm. All rights reserved.
+//
+
+import Combine
+
+extension Publisher {
+    /// Converts any publisher to a future.
+    ///
+    /// - returns: A publisher that eventually produces a single value and then finishes or fails.
+    func asFuture() -> Publishers.Future<Self> {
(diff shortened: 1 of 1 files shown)


freak4pc:
Hey @jullianm, hey there :) 
This is an interesting idea but there are a few issues to discuss here:

1. You need to add tests, documentation, etc. 
2. You're not creating a Combine Future, but instead you're creating your own Single-valued-publisher. This is very problematic because Future has specific traits (it's greedy, it shares a value replay to future subscribers, etc). This is indeed more similar to Single, which is vastly different compared to Combine's Future. 

If traits are what you're interested in, there's actually a branch called `traits` where I started implementing Completable and Single. I plan to upload it soon :) 

Thanks again for the contribution!

jullianm:
Hey Shai,

Thanks for the quick feedbacks.
I didn’t know about these differences between Future and Single so thanks for the clarification!

I look forward to seeing the traits branch :)

Jullian

freak4pc:
Cool! Should be in a couple of days :) 
I'll be closing this for now.

PR #14, opened 2020-04-03
Adds `Publisher.weaklyAssign(to:on:)`.

jasdev:
Would love any input on naming here—no strong opinions on `.weaklyAssign`. 👌🏽

Also, GitHub’s parsing of my commit message. 😂

Screen Shot 2020-04-02 at 9 45 32 PM

### DIFF
--- a/CombineExt.xcodeproj/project.pbxproj
+++ b/CombineExt.xcodeproj/project.pbxproj
@@ -41,6 +41,8 @@
 		78C193E0241D4D8D0001B7FD /* MaterializeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 78C193DF241D4D8D0001B7FD /* MaterializeTests.swift */; };
 		78C193E2241D596F0001B7FD /* Dematerialize.swift in Sources */ = {isa = PBXBuildFile; fileRef = 78C193E1241D596F0001B7FD /* Dematerialize.swift */; };
 		78C193E4241D63620001B7FD /* DematerializeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 78C193E3241D63620001B7FD /* DematerializeTests.swift */; };
+		AAEAF0E32436C8A8007C35E0 /* WeakAssign.swift in Sources */ = {isa = PBXBuildFile; fileRef = AAEAF0E22436C8A8007C35E0 /* WeakAssign.swift */; };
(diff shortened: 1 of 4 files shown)


freak4pc:
Interesting idea and quite small as well, I'm only trying to think if it's actually useful enough - I'm worried of adding too many "novelty" operators. 

I don't want tens of operators called `mapWeakly`, `sinkWeakly`, etc. Maybe there's a nice way to generalize it?

sharplet:
Naming-wise, I'm not sure `weaklyAssign`, `weakAssign`, etc. are quite right, because it's the object reference that's weak, not the assignment. Following that logic, if feels like the modifier belongs on the `object` parameter, so perhaps something like this?

```swift
publisher.assign(to: \.foo, onWeak: bar)`
```

This mirrors capture list syntax somewhat, where it would be spelled `[weak bar]`. It also perhaps lends itself to this variant for `unowned`:

```swift
publisher.assign(to: \.foo, onUnowned: bar)
```

I'm sympathetic to @freak4pc's point about generality though. Another option I considered was a function that returns a weak reference, e.g.:

```swift
final class WeakBox<Object: AnyObject> {
  weak var object: Object?

  init(_ object: Object) {
    self.object = object
  }
}

func weak<Object: AnyObject>(_ object: Object) -> WeakBox<Object> {
  WeakBox(object)
}

publisher.assign(to: \.foo, on: weak(bar))
```

I don't love it, but I don't hate it. In [Bindings](https://github.com/thoughtbot/Bindings) I sidestepped this issue altogether by introducing the `BindingOwner` protocol, which defines weak capture semantics at the type level, rather than in an ad-hoc fashion through operator overloads.

jasdev:
@freak4pc, @sharplet: Thank you both for the feedback, means a lot that y’all took the time. 🙏🏽

And totally agree on the ad-hoc’ness of `Publisher.weaklyAssign(to:on)`. I’m guessing the `Bindings` approach is a bit out of scope for CombineExt—though, Shai can better speak to that 😄. For now, I’ll close this and roll with `weak`-as-a-function in my projects, since I’m digging how it sort of “promotes” the keyword to a function.

Cheers! 🍻

jasdev:
(For any future souls stumbling upon this, trying to make `WeakBox` and `weak` (the free function) work [taught me way more than I bargained for](https://iosdevelopers.slack.com/archives/C03FWU59B/p1586378364117100) about dynamic member lookup. 😂)

PR #17, opened 2020-04-07
Add `tryMapMany(_:)` operator. 

jdisho:
Adding the version of [`mapMany`](https://github.com/CombineCommunity/CombineExt/pull/9) with a throwable `transform` closure. 🎉

**Example:** 

```swift 
let intArrayPublisher = PassthroughSubject<[Int], NumberError>()
    
intArrayPublisher
    .tryMapMany { number in
        if number < 5 {
            throw NumberError.numberSmallerThanFive
        }
        return "\(number)"
    }
    .sink(receiveValue: { print($0) })

intArrayPublisher.send([10, 20, 20, 40, 30, 8])
```

**Output:**
```none
["10", "20", "20", "40", "30", "8"]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ All operators, utilities and helpers respect Combine's publisher contract, inclu
 * [zip(with:) and Collection.zip](#ZipMany)
 * [combineLatest(with:) and Collection.combineLatest](#CombineLatestMany)
 * [mapMany(_:)](#MapMany)
+* [tryMapMany(_:)](#TryMapMany)
 * [setOutputType(to:)](#setOutputType)
 
 ### Publishers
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -363,6 +364,32 @@ intArrayPublisher.send([10, 2, 2, 4, 3, 8])
 ["10", "2", "2", "4", "3", "8"]
 ```
 
+### TryMapMany
+
+Projects each element of a publisher collection into a new publisher collection form.
```
freak4pc, line 369: I would just merge this documentation with the `mapMany` one mentioning that there's a try-prefixed variant from which you can throw an error and show this code example

WDYT?

Sources/Operators/TryMapMany.swift
```diff
@@ -0,0 +1,55 @@
...
+    ///    // Output: ["10", "2", "2", "4", "3", "8"]
+    ///    ```
+    ///
+    /// In case of an error:
+    ///
+    ///     let intArrayPublisher = PassthroughSubject<[Int], NumberError>()
+    ///
+    ///     intArrayPublisher
+    ///         .tryMapMany { number in
+    ///             if number < 5 {
```
freak4pc, line 40: just nitpicky but this wont compile - you're comparing to a number but `number` is an array of integers

jdisho: 🤪

jdisho:
Omg, I deleted the old fork and now I cannot do changes to this PR. 
I assume I have to open a new one 😔

Or do you know any other way?

freak4pc:
I've pushed it to a branch here, seems like it does the trick lol: 
https://github.com/CombineCommunity/CombineExt/tree/try-map-many

Sources/Operators/TryMapMany.swift
jdisho: I got a bit confused, but the `number`'s type is an `Int` and not `[Int]`.

PR #21, opened 2020-04-11
`Set.store(_:)` `AnyCancellable`-constrained storage overloads.

jasdev:
Figured it’d be helpful to have an anlog to the [`DisposeBag.insert(_:)` variants](https://github.com/ReactiveX/RxSwift/blob/002d325b0bdee94e7882e1114af5ff4fe1e96afa/RxSwift/Disposables/DisposeBag.swift#L100-L113).

### DIFF
--- a/CombineExt.xcodeproj/project.pbxproj
+++ b/CombineExt.xcodeproj/project.pbxproj
@@ -43,6 +43,8 @@
 		78C193E0241D4D8D0001B7FD /* MaterializeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 78C193DF241D4D8D0001B7FD /* MaterializeTests.swift */; };
 		78C193E2241D596F0001B7FD /* Dematerialize.swift in Sources */ = {isa = PBXBuildFile; fileRef = 78C193E1241D596F0001B7FD /* Dematerialize.swift */; };
 		78C193E4241D63620001B7FD /* DematerializeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 78C193E3241D63620001B7FD /* DematerializeTests.swift */; };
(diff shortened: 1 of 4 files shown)


freak4pc:
Hey @jasdev - thanks for this :) In this specific case though, I think this PR doesn't really fit the spirit of CombineExt, because this extension effects `Set` as a whole and only allows really light ergonomics. 

The more "generic" version of this would allow `set.insert(contentsOf: [...])`, which is a behavior you can get if you use an `Array<AnyCancellable>`, instead.

I don't think this specific extension is useful enough to include here and am more worried about the fact it extends a Foundation type, and not a Combine type. 

WDYT?

jasdev:
> because this extension effects `Set` as a whole

For what it’s worth, it is constrained to `Set.Element == AnyCancellable`. So, its surface area is as small as possible.

> which is a behavior you can get if you use an `Array<AnyCancellable>`, instead.

Yep! Multi-inserting isn’t available on `Set` and I figured the cosmetic is worth it so folks don’t have to fall back for an `Array` when a `Set` might be more performant in handling accidental double-inserts of the same cancellable.

>  it extends a Foundation type, and not a Combine type.

While true, it again is constrained to a Combine type. So it’s only surfaced to folks working with `Set<AnyCancellable>`s.

---

I don’t have a strong opinion here though, so, if you’re not digging it I’m totally cool closing this. 😄

freak4pc:
It feels wrong to add, in my eyes. I might've extended this on a personal project but I'm feeling weird adding an extension on Set for a project around Combine extensions and operators, especially since it's quite simple and would be trivial to write yourself if its ergonomics you're interested in having.

I think that the way to go would be to find more people who think this is a good addition to have as part of CombineExt, in which case we can add this, but otherwise I'd prefer to leave it out.

jasdev:
10-4!

PR #22, opened 2020-04-11
🎉 `withLatestFrom` for more than one publisher.

jdisho:
I'm working on creating `withLatestFrom` operators that support more than one publisher.
In RxSwift, I face many scenarios where I need to get the latest values from more than one observable and additionally there is this great [thread](https://github.com/ReactiveX/RxSwift/issues/1973).

Since under the hood, I'm using `combineLatest`, I'm supporting as many publishers as the operator supports (up to four).

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ This section outlines some of the custom operators CombineExt provides.
 
 ### withLatestFrom
 
-Merges two publishers into a single publisher by combining each value from `self` with the _latest_ value from the second publisher, if any.
+Merges up to fours publishers into a single publisher by combining each value from `self` with the _latest_ value from the other publishers, if any.
 
 ```swift
 let taps = PassthroughSubject<Void, Never>()
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ This section outlines some of the custom operators CombineExt provides.
 
 ### withLatestFrom
 
-Merges up to fours publishers into a single publisher by combining each value from `self` with the _latest_ value from the other publishers, if any.
+Merges up to four publishers into a single publisher by combining each value from `self` with the _latest_ value from the other publishers, if any.
 
 ```swift
 let taps = PassthroughSubject<Void, Never>()
--- a/Sources/Operators/WithLatestFrom.swift
+++ b/Sources/Operators/WithLatestFrom.swift
@@ -13,7 +13,7 @@ public extension Publisher {
   ///  Merges two publishers into a single publisher by combining each value
   ///  from self with the latest value from the second publisher, if any.
   ///
-  ///  - parameter other: Second observable source.
+  ///  - parameter other: A second publisher source.
   ///  - parameter resultSelector: Function to invoke for each value from the self combined
   ///                              with the latest value from the second source, if any.
   ///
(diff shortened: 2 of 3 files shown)


freak4pc:
Hey @jdisho, first of all thank you so much for your work.

TBH I have some issues with adding something like this:

1. I agree with the thread you linked that this should be a per-project/repo thing because it is a specific constraint of _some_, but not _all_, use cases.

2. The bigger issue is that this tucks in (e.g hides) the fact combineLatest is used under the hood which usually has no weird side-effects but it still quite strange. 

I’m all for keeping this issue open to see if there’s more practical interest in adding this, and also to think a bit more if this is a good fit for this library.

Makes sense to you? WDYT?

jdisho:
Hey @freak4pc, thanks for your explanation. We can leave it open and see if other folks have different opinions. 😄

However, can you elaborate a bit more on your points? B/c, I don't fully understand them. 😇

1. >...  because it is a specific constraint of some, but not all, use cases. 

     What can be some other cases where the implementation with `combineLatest` can be    incorrect or have unexpected side effects? 
    
     In an example, don't you **always** expect on `button` tap, to get the latest value from `publisher1` **and** `publisher2`?
     ```swift 
     button.tapPublisher.withLatestFrom(publisher1, publisher2)
    ```

2. >... the fact combineLatest is used under the hood which usually has no weird side-effects but it still quite strange.
   
    Why do you think this is strange? Is this just your personal opinion? Based on this argument, every new operator that reuses given ones under the hood can feel strange. 🙇‍♂️

jdisho:
What does @jasdev think?

freak4pc:
1. Yes, some people don't have the use case of combining multiple inputs this way on a button tap. Also the point the original thread makes is that this is very simplistic sugar that is perhaps left explicit given it doesn't provide any "logic" but just basically removes "CombineLatest" from the call-site. These sugar-y operators (IMO) do require thinking over more carefully if they're actually needed. 

2. Of course it's my personal opinion, I can't have anyone else's opinion but my own :)

jdisho:
IMO, the user of this operator shouldn't care about the implantation details. I think what's important is the operator's description to match the expectations. 

If this is an option, we can create `WithLatestFrom2` and `WithLatestFrom3` without using `combineLatest`.

freak4pc:
Ok, I thought of it some more, I think this could be an OK addition. I'd be happy to get more opinions though.

Sources/Operators/WithLatestFrom.swift
```diff
@@ -36,6 +36,43 @@ public extension Publisher {
...
+  /// - returns: A publisher containing the latest value from the second and third publisher, if any.
+  func withLatestFrom<Other, Other1>(_ other: Other,
+                                     _ other1: Other1)
+    -> AnyPublisher<(Self.Output, Other.Output, Other1.Output), Failure>
+    where Other: Publisher, Other1: Publisher, Other.Failure == Failure, Other1.Failure == Failure {
+      let combined = combineLatest(other, other1)
+        .eraseToAnyPublisher()
+      return withLatestFrom(combined)
+        .eraseToAnyPublisher()
+    }
```
freak4pc, line 55: Can we avoid erasing to have a `WithLatestFrom<AnyPublisher` somehow?

freak4pc, line 55: I think we can go down this road:

```suggestion
  /// Upon an emission from self, emit the latest value from the
  /// second and third publisher, if any exists.
  ///
  /// - parameter other: A second publisher source.
  /// - parameter other1: A third publisher source.
  ///
  /// - returns: A publisher containing the latest value from the second and third publisher, if any.
  func withLatestFrom<Other: Publisher, Other1: Publisher>(_ other: Other,
                                                           _ other1: Other1)
      -> Publishers.WithLatestFrom<Self,
                                   AnyPublisher<(Other.Output, Other1.Output), Self.Failure>,
                                   (Self.Output, Other.Output, Other1.Output)>
      where Other.Failure == Failure, Other1.Failure == Failure {
          let combined = other.combineLatest(other1)
              .eraseToAnyPublisher()

          return withLatestFrom(combined, resultSelector: { ($0, $1.0, $1.1) })
  }
```

Sources/Operators/WithLatestFrom.swift
```diff
@@ -22,19 +22,94 @@ public extension Publisher {
...
+  ///
+  /// - parameter other: A second publisher source.
+  /// - parameter other1: A third publisher source.
+  ///
+  /// - returns: A publisher containing the latest value from the second and third publisher, if any.
+  func withLatestFrom<Other: Publisher, Other1: Publisher>(_ other: Other,
+                                                           _ other1: Other1)
+    -> Publishers.WithLatestFrom<Self, AnyPublisher<(Other.Output, Other1.Output), Self.Failure>, (Other.Output, Other1.Output)>
+    where Other.Failure == Failure, Other1.Failure == Failure {
+      return withLatestFrom(other, other1, resultSelector: { $1 })
```
freak4pc, line 96:
```suggestion
     withLatestFrom(other, other1) { $1 }
```

Sources/Operators/WithLatestFrom.swift
```diff
@@ -22,19 +22,94 @@ public extension Publisher {
...
+  /// - parameter other1: A third publisher source.
+  /// - parameter other2: A forth publisher source.
+  ///
+  /// - returns: A publisher containing the latest value from the second, third and forth publisher, if any.
+  func withLatestFrom<Other: Publisher, Other1: Publisher, Other2: Publisher>(_ other: Other,
+                                                                              _ other1: Other1,
+                                                                              _ other2: Other2)
+    -> Publishers.WithLatestFrom<Self, AnyPublisher<(Other.Output, Other1.Output, Other2.Output), Self.Failure>, (Other.Output, Other1.Output, Other2.Output)>
+    where Other.Failure == Failure, Other1.Failure == Failure, Other2.Failure == Failure {
+      return withLatestFrom(other, other1, other2,  resultSelector: { $1 })
```
freak4pc, line 112:
```suggestion
     withLatestFrom(other, other1, other2) { $1 }
```

jdisho:
Always forget that! 😄
Yes, I will take care. Thanks for the review.

freak4pc:
Do you have time to finish the README here? I'm happy to wrap it for you. Let me know ;)

jdisho:
Working on it now 🤗

README.md
```diff
@@ -74,9 +74,11 @@ github "CombineCommunity/CombineExt"
 
 This section outlines some of the custom operators CombineExt provides.
 
-### withLatestFrom
+## withLatestFrom
```
freak4pc, line 77: oops?

README.md
```diff
@@ -74,9 +74,11 @@ github "CombineCommunity/CombineExt"
...
 
-Merges two publishers into a single publisher by combining each value from `self` with the _latest_ value from the second publisher, if any.
+### withLatestFrom(_:)
+
+Merges **two** publishers into a single publisher by combining each value from `self` with the _latest_ value from the second publisher, if any.
```
freak4pc, line 81: Let's not add specific documentation per-overload:) 
We should just expand withLatestFrom mentioning it supports overloads of up to 4 observables

jdisho: I updated te README based on this comment and I hope I got it right. Should it be fine now? 😄

README.md
```diff
@@ -76,7 +76,7 @@ This section outlines some of the custom operators CombineExt provides.
 
 ### withLatestFrom
 
-Merges two publishers into a single publisher by combining each value from `self` with the _latest_ value from the second publisher, if any.
+Merges up to fours publishers into a single publisher by combining each value from `self` with the _latest_ value from the other publishers, if any.
```
freak4pc, line 79:
```suggestion
Merges up to four publishers into a single publisher by combining each value from `self` with the _latest_ value from the other publishers, if any.
```

🤭

PR #23, opened 2020-04-13
Adds ReplaySubject and Publisher.share(replay:)

jasdev:
h/t to @tcldr for the reference implementation I leaned on and adapted to Shai’s `DemandBuffer` type. 💯

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ All operators, utilities and helpers respect Combine's publisher contract, inclu
 * [mapMany(_:)](#MapMany)
 * [setOutputType(to:)](#setOutputType)
 * [removeAllDuplicates and removeAllDuplicates(by:) ](#removeAllDuplicates)
-* [share(replay:)](#ShareReplay)
+* [share(replay:)](#share(replay:))
 
 ### Publishers
 * [AnyPublisher.create](#AnypublisherCreate)
@@ -438,7 +438,7 @@ removeAllDuplicates: 4
 
 ------
 
-### ShareReplay
+### share(replay:)
 
 Similar to [`Publisher.share`](https://developer.apple.com/documentation/combine/publisher/3204754-share), `.share(replay:)` can be used to create a publisher instance with reference semantics which replays a pre-defined amount of value events to further subscribers.
 
@@ -579,7 +579,7 @@ guarantees
 A Combine analog to Rx’s [`ReplaySubject` type](http://reactivex.io/documentation/subject.html). It’s similar to a [`CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject) in that it buffers values, but, it takes it a step further in allowing consumers to specify the number of values to buffer and replay to future subscribers. Also, it will handle forwarding any completion events after the buffer is cleared upon subscription.
 
 ```swift
-let subject = ReplaySubject<Int, Never>(maxBufferSize: 3)
+let subject = ReplaySubject<Int, Never>(bufferSize: 3)
 
 subject.send(1)
 subject.send(2)
--- a/README.md
+++ b/README.md
@@ -440,7 +440,7 @@ removeAllDuplicates: 4
 
 ### ShareReplay
 
-Similar to [`Publisher.share`](https://developer.apple.com/documentation/combine/publisher/3204754-share), `.share(replay:)` can be used to create a publisher instance with reference semantics. But, it also replays value events to further subscribers.
+Similar to [`Publisher.share`](https://developer.apple.com/documentation/combine/publisher/3204754-share), `.share(replay:)` can be used to create a publisher instance with reference semantics which replays a pre-defined amount of value events to further subscribers.
 
 ```swift
 let subject = PassthroughSubject<Int, Never>()
--- a/README.md
+++ b/README.md
@@ -576,7 +576,7 @@ guarantees
 
 ### ReplaySubject
 
-A Combine analog to Rx’s [`ReplaySubject` type](http://reactivex.io/documentation/subject.html). It’s similar to a [`CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject) in that it buffers elements, but, it takes it a step further in allowing consumers to specific the number of value events to buffer and replay to future subscribers. Also, it will handle forwarding any completion events after the buffer is cleared upon subscription.
+A Combine analog to Rx’s [`ReplaySubject` type](http://reactivex.io/documentation/subject.html). It’s similar to a [`CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject) in that it buffers values, but, it takes it a step further in allowing consumers to specify the number of values to buffer and replay to future subscribers. Also, it will handle forwarding any completion events after the buffer is cleared upon subscription.
 
 ```swift
 let subject = ReplaySubject<Int, Never>(maxBufferSize: 3)
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ All operators, utilities and helpers respect Combine's publisher contract, inclu
 * [mapMany(_:)](#MapMany)
 * [setOutputType(to:)](#setOutputType)
 * [removeAllDuplicates and removeAllDuplicates(by:) ](#removeAllDuplicates)
-* [share(replay:)](#share(replay:))
+* [share(replay:)](#sharereplay)
 
 ### Publishers
 * [AnyPublisher.create](#AnypublisherCreate)
--- a/README.md
+++ b/README.md
@@ -34,11 +34,15 @@ All operators, utilities and helpers respect Combine's publisher contract, inclu
 * [mapMany(_:)](#MapMany)
 * [setOutputType(to:)](#setOutputType)
 * [removeAllDuplicates and removeAllDuplicates(by:) ](#removeAllDuplicates)
+* [share(replay:)](#ShareReplay)
 
 ### Publishers
 * [AnyPublisher.create](#AnypublisherCreate)
 * [CurrentValueRelay](#CurrentValueRelay)
 * [PassthroughRelay](#PassthroughRelay)
+
+### Subjects
+* [ReplaySubject](#ReplaySubject)
 
 > **Note**: This is still a relatively early version of CombineExt, with much more to be desired. I gladly accept PRs, ideas, opinions, or improvements. Thank you! :)
 
@@ -173,6 +177,8 @@ amb: 3
 amb: 6
 amb: completed with .finished
 ```
+
+------
 
 ### materialize
 
@@ -348,6 +354,8 @@ You may also use `.zip()` directly on a collection of publishers with the same o
 zipped: 10
 ```
 
+------
+
 ### CombineLatestMany
 
 This repo includes two overloads on Combine’s `Publisher.combineLatest` methods (which, at the time of writing only go up to arity three) and an `Collection.combineLatest` constrained extension.
@@ -379,6 +387,8 @@ combineLatest: [true, true, true, true]
 combineLatest: [false, true, true, true]
 ```
 
+------
+
 ### MapMany
 
 Projects each element of a publisher collection into a new publisher collection form.
@@ -399,9 +409,13 @@ intArrayPublisher.send([10, 2, 2, 4, 3, 8])
 ["10", "2", "2", "4", "3", "8"]
 ```
 
+------
+
 ### setOutputType
 
 `Publisher.setOutputType(to:)` is an analog to [`.setFailureType(to:)`](https://developer.apple.com/documentation/combine/publisher/3204753-setfailuretype) for when `Output` is constrained to `Never`. This is especially helpful when chaining operators after an [`.ignoreOutput()`](https://developer.apple.com/documentation/combine/publisher/3204714-ignoreoutput) call.
+
+------
 
 ### removeAllDuplicates
 
@@ -420,6 +434,40 @@ removeAllDuplicates: 1
 removeAllDuplicates: 2
 removeAllDuplicates: 3
 removeAllDuplicates: 4
+```
+
+------
+
+### ShareReplay
+
+Similar to [`Publisher.share`](https://developer.apple.com/documentation/combine/publisher/3204754-share), `.share(replay:)` can be used to create a publisher instance with reference semantics. But, it also replays value events to further subscribers.
+
+```swift
+let subject = PassthroughSubject<Int, Never>()
+
+let replayedPublisher = subject
+  .share(replay: 3)
+
+subscription1 = replayedPublisher
+  .sink(receiveValue: { print("first subscriber: \($0)") })
+  
+subject.send(1)
+subject.send(2)
+subject.send(3)
+subject.send(4)
+
+subscription2 = replayedPublisher
+  .sink(receiveValue: { print("second subscriber: \($0)") })
+```
+
+```none
+first subscriber: 1
+first subscriber: 2
+first subscriber: 3
+first subscriber: 4
+second subscriber: 2
+second subscriber: 3
+second subscriber: 4
 ```
 
 ## Publishers
@@ -522,6 +570,35 @@ only
 provide
 great
 guarantees
+```
+
+## Subjects
+
+### ReplaySubject
+
+A Combine analog to Rx’s [`ReplaySubject` type](http://reactivex.io/documentation/subject.html). It’s similar to a [`CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject) in that it buffers elements, but, it takes it a step further in allowing consumers to specific the number of value events to buffer and replay to future subscribers. Also, it will handle forwarding any completion events after the buffer is cleared upon subscription.
+
+```swift
+let subject = ReplaySubject<Int, Never>(maxBufferSize: 3)
+
+subject.send(1)
+subject.send(2)
+subject.send(3)
+subject.send(4)
+
+subject
+  .sink(receiveValue: { print($0) })
+
+subject.send(5)
+```
+
+#### Output:
+
+```none
+2
+3
+4
+5
 ```
 
 ## License
--- a/Sources/Operators/ShareReplay.swift
+++ b/Sources/Operators/ShareReplay.swift
@@ -1,0 +1,21 @@
+//
+//  ShareReplay.swift
+//  CombineExt
+//
+//  Created by Jasdev Singh on 13/04/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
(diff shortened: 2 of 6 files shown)


Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,179 @@
...
+        // MARK: - Internal
+
+        let innerSubscriberIdentifier: CombineIdentifier
+
+        init(
+            downstream: Downstream,
+            cancelHandler: (() -> Void)?
+        ) {
+            self.sink = Sink(
+                upstream: Empty(completeImmediately: false), // `ReplaySubject`’s `Subject` conformance handles
```
jasdev, line 130: I was a bit unsure how to handle not having a “real” upstream, so, I’d love any thoughts, if I’m misusing `Sink` here by passing an `Empty` upstream (and instead leaning on `ReplaySubject`’s `Subject` requirements to forward on events).

tcldr:
> h/t to @tcldr for the reference implementation I leaned on and adapted to Shai’s `Sink` type. 💯 

Glad it’s been useful! 👍

README.md
```diff
@@ -422,6 +436,40 @@ removeAllDuplicates: 3
 removeAllDuplicates: 4
 ```
 
+------
+
+### ShareReplay
```
freak4pc, line 441:
```suggestion
### share(replay:)
```

README.md
```diff
@@ -422,6 +436,40 @@ removeAllDuplicates: 3
...
+
+Similar to [`Publisher.share`](https://developer.apple.com/documentation/combine/publisher/3204754-share), `.share(replay:)` can be used to create a publisher instance with reference semantics. But, it also replays value events to further subscribers.
```
freak4pc, line 443:
```suggestion
Similar to [`Publisher.share`](https://developer.apple.com/documentation/combine/publisher/3204754-share), `.share(replay:)` can be used to create a publisher instance with reference semantics which replays a pre-defined amount of value events to further subscribers.
```

README.md
```diff
@@ -518,6 +566,35 @@ great
 guarantees
 ```
 
+## Subjects
+
+### ReplaySubject
+
+A Combine analog to Rx’s [`ReplaySubject` type](http://reactivex.io/documentation/subject.html). It’s similar to a [`CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject) in that it buffers elements, but, it takes it a step further in allowing consumers to specific the number of value events to buffer and replay to future subscribers. Also, it will handle forwarding any completion events after the buffer is cleared upon subscription.
```
freak4pc, line 573:
```suggestion
A Combine analog to Rx’s [`ReplaySubject` type](http://reactivex.io/documentation/subject.html). It’s similar to a [`CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject) in that it buffers values, but, it takes it a step further in allowing consumers to specify the number of values to buffer and replay to future subscribers. Also, it will handle forwarding any completion events after the buffer is cleared upon subscription.
```

README.md
```diff
@@ -518,6 +566,35 @@ great
...
+
+```swift
+let subject = ReplaySubject<Int, Never>(maxBufferSize: 3)
```
freak4pc, line 576: Any reason for the `max`? I'd just go `bufferSize`

Sources/Operators/ShareReplay.swift
```diff
@@ -0,0 +1,21 @@
...
+//  CombineExt
+//
+//  Created by Jasdev Singh on 13/04/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import Combine
+
+public extension Publisher {
+    /// An overload on [Publisher.share](https://developer.apple.com/documentation/combine/publisher/3204754-share)
```
freak4pc, line 12: Maybe instead of overload we should just say it's a variation of the share operator? 
I'd also remove the links from the documentation TBH since Xcode doesn't actually knows how to parse these and they create a bit of noise when trying to read the notes (same for other pieces of comments). WDYT?

Sources/Operators/ShareReplay.swift
```diff
@@ -0,0 +1,21 @@
...
+    /// that allows for buffering and replaying a `replay` amount of value events to future subscribers.
+    /// - Parameter count: The number of value events to buffer in a [first-in-first-out](https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics)) manner.
+    /// - Returns: A type-erased, shared publisher that replays the specified number of value events to future subscribers.
+    func share(replay count: UInt) -> AnyPublisher<Output, Failure> {
```
freak4pc, line 16: Would use an `Int` here, some people might use a constant for the replay and I wouldn't want them to need to cast. I don't think there's any dramatic performance benefits to doing this here

Sources/Operators/ShareReplay.swift
```diff
@@ -0,0 +1,21 @@
...
+        multicast { ReplaySubject(maxBufferSize: count) }
```
freak4pc, line 17: Might be worth specializing for 1 with CurrentValueSubject and over 1 with ReplaySubject ? 

```suggestion
        multicast { ReplaySubject(maxBufferSize: count) }
```

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+//  CombineExt
+//
+//  Created by Jasdev Singh on 13/04/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import Combine
+
+/// `ReplaySubject` is a [CurrentValueSubject](https://developer.apple.com/documentation/combine/currentvaluesubject)
+/// that can buffer, well, more than one value! It stores value events, up to its `maxBufferSize` in a
```
freak4pc, line 12: I do like a more comical writing vibe for blogs, but I'm not sure it fits as much for code documentation. Happy for your thoughts about this 
```suggestion
/// that can buffer more than one value! It stores value events, up to its `maxBufferSize` in a
```

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+/// [first-in-first-out](https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics)) manner and then replays it to
+/// future subscribers and also forwards completion events.
+///
+/// The implementation borrows heavily from [Tristan’s](https://github.com/tcldr/Entwine/blob/b839c9fcc7466878d6a823677ce608da998b95b9/Sources/Entwine/Operators/ReplaySubject.swift)
+/// with the main modifications being leaning on Shai’s [DemandBuffer](https://github.com/CombineCommunity/CombineExt/blob/7ed4677e33fdc963af404423cb563e133c271d2b/Sources/Common/DemandBuffer.swift)
```
freak4pc, line 17:
```suggestion
/// with the main modifications being leaning on CombineExt's [DemandBuffer](https://github.com/CombineCommunity/CombineExt/blob/7ed4677e33fdc963af404423cb563e133c271d2b/Sources/Common/DemandBuffer.swift)
```

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+/// instead of `Entwine`’s [SinkQueue](https://github.com/tcldr/Entwine/blob/b839c9fcc7466878d6a823677ce608da998b95b9/Sources/Common/Utilities/SinkQueue.swift)
+/// and a plain ol’ `[Output]` buffer instead of a
+/// [custom, LinkedListQueue-backed one](https://github.com/tcldr/Entwine/blob/8be24a59bc91410bb29e84b1c4ae35398a5839c8/Sources/Entwine/Operators/ReplaySubject.swift#L162-L177).
```
freak4pc, line 20: I'm all for sharing the credit but I'm not sure detailing the differences belongs in a class documentation, just my two cents. People read the docs to understand how the class works, not really on how it does or doesn't differ from its source of influence. 

I'd probably leave just this: 

> The implementation borrows heavily from [Entwine's ReplaySubject](https://github.com/tcldr/Entwine/blob/b839c9fcc7466878d6a823677ce608da998b95b9/Sources/Entwine/Operators/ReplaySubject.swift)

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+public final class ReplaySubject<Output, Failure: Error> {
+    // MARK: - Private
+
```
freak4pc, line 23:
```suggestion
```

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+    private let maxBufferSize: UInt
+    private var buffer = [Output]()
+
+    // MARK: - Internal
+
```
freak4pc, line 28:
```suggestion
```

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+    var subscriptions = [Subscription<AnySubscriber<Output, Failure>>]()
```
freak4pc, line 29: Might be worth adding a comment about why the individual subscriptions are stored (I imagine it's to be able to cast the values to everyone)

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+
+    // We also track subscriber identifiers, to more quickly bottom-out double subscribes instead of having to do a
+    // linear pass over `subscriptions`.
+    var subscriberIdentifiers = Set<CombineIdentifier>()
+
+    private var completion: Subscribers.Completion<Failure>?
+    private var isActive: Bool { completion == nil }
+
+    /// The initialization point for `ReplaySubject`s.
```
freak4pc, line 38: perhaps: 
```suggestion
    /// Create a `ReplaySubject` buffering up to `maxBufferSize` values and replaying them to new subscribers
```

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+    /// - Parameter maxBufferSize: The maximum number of value events to buffer and replay to all future subscribers.
+    public init(maxBufferSize: UInt) {
```
freak4pc, line 40: same thing with the UInt, don't see a value in this (would use Int across the board for the sake of ergonomics)

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+    public typealias Failure = Failure
+
+    public func receive<Subscriber: Combine.Subscriber>(
+        subscriber: Subscriber
+    ) where Failure == Subscriber.Failure, Output == Subscriber.Input {
+        let subscriberIdentifier = subscriber.combineIdentifier
+
+        guard !subscriberIdentifiers.contains(subscriberIdentifier) else {
+            subscriber.receive(subscription: Subscriptions.empty)
+            subscriber.receive(completion: .finished)
```
freak4pc, line 86: is it correct sending a finishing event here?

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        let subscription = Subscription(
+            downstream: AnySubscriber(subscriber)
+        ) { [weak self] in
+            guard
+                let self = self,
+                let subscriptionIndex = self.subscriptions
+                    .firstIndex(where: { $0.innerSubscriberIdentifier == subscriberIdentifier })
+                else { return }
```
freak4pc, line 97:
```suggestion
        let subscription = Subscription(
            downstream: AnySubscriber(subscriber)) { [weak self] in
            guard let self = self,
                  let subscriptionIndex = self.subscriptions
                                              .firstIndex(where: { $0.innerSubscriberIdentifier == subscriberIdentifier }) else { return }
```

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+        subscriber.receive(subscription: subscription)
+        subscription.replay(buffer, completion: completion)
+    }
+}
+
+// MARK: - `ReplaySubject.Subscription`
+
+extension ReplaySubject {
+    final class Subscription<Downstream: Subscriber>
+    where Output == Downstream.Input, Failure == Downstream.Failure {
```
freak4pc, line 115: prefer to one-line this like other publishers in this repo

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,171 @@
...
+        private var cancelHandler: (() -> Void)?
+
+        // MARK: - Internal
+
+        let innerSubscriberIdentifier: CombineIdentifier
+
+        init(
+            downstream: Downstream,
+            cancelHandler: (() -> Void)?
+        ) {
```
freak4pc, line 128:
```suggestion
        init(downstream: Downstream, cancelHandler: (() -> Void)?) {
```

Might also be worth naming cancelationHandler, but up to you

Tests/ReplaySubjectTests.swift
```diff
@@ -0,0 +1,358 @@
...
+import Combine
+@testable import CombineExt
+import XCTest
+
+final class ReplaySubjectTests: XCTestCase {
+    private var cancellable1: AnyCancellable!
+    private var cancellable2: AnyCancellable!
+    private var cancellable3: AnyCancellable!
+
+    private enum  AnError: Error, Equatable {
```
freak4pc, line 17:
```suggestion
    private enum AnError: Error, Equatable {
```

Sources/Operators/ShareReplay.swift
freak4pc: Also, could be cool to return `Publishers.Multicast` instead of erasing the entire thing (while erasing the inner-constraints of the multi-cast, perhaps)

Sources/Subjects/ReplaySubject.swift
freak4pc: Also saying "ReplaySubject is a CurrentValueSubject" might be misleading. WDYT about "ReplaySubject is a subject capable of buffering more than a single value" 

I also think comparing to CurrentValue might be confusing to people who would actually expect to do `.value` on it

Sources/Operators/ShareReplay.swift
jasdev: Seems to parse fine, at least client-side. I’m a fan of linking out, but, if they’re too noisy in your opinion, I’m game to remove them.

Screen Shot 2020-04-18 at 8 21 31 PM

README.md
jasdev: Updating in the next commit. 👌🏽

Sources/Operators/ShareReplay.swift
jasdev: Sure thing—swapping in an `Int` in the next commit.

Sources/Operators/ShareReplay.swift
jasdev: > Might be worth specializing for 1 with CurrentValueSubject

Hah! So part of my next blog post digs into this—it’s subtle. But, we can’t lean on `CurrentValueSubject` since it behaves like a `BehaviorSubject` and doesn’t replay any values post-terminal event.

And game to return a `Publishers.Multicast` with an erased inner. Incoming in the next commit.

Sources/Subjects/ReplaySubject.swift
jasdev: mm yea, actually wondering about this too now (cc @tcldr)—when double subscribing on a `Passthrough-` or `CurrentValueSubject`, I’m not seeing any added terminal events. So, I’m guessing passing down an empty subscription might be all that we should do, too.

Sources/Subjects/ReplaySubject.swift
jasdev: Agreed, re-worded. 💯

jasdev:
Should be good for another pass over, thanks for the thorough review, Shai. 🙏🏽

Sources/Operators/ShareReplay.swift
freak4pc: Ah cool! Probably some of the new markdown features :) 
I'd say - leave links that are references to other important code-things like `share()`, but I'd remove links to educational non-code content (like FIFO on Wikipedia) .

Also in this specific case might be worth saying 'A variation on `share()`, since it's a documentation on the operator iself (the method) and not the publisher  (concrete type)

Sources/Operators/ShareReplay.swift
freak4pc: That's interesting :) Well, we have BehaviorRelay in the repo which would help with that, but sure we can leave it as is :] thanks!

Sources/Subjects/ReplaySubject.swift
tcldr: Yes, just confirmed this myself – thanks for the heads-up! Certainly best to try and maintain consistency with the behaviour that Apple has implemented in their own Subjects.

jasdev:
Aaand round-two feedback incorporated. 🤞🏽

Thanks again for the reads over this!

Sources/Subjects/ReplaySubject.swift
```diff
@@ -0,0 +1,129 @@
...
+    public func receive<Subscriber: Combine.Subscriber>(subscriber: Subscriber) where Failure == Subscriber.Failure, Output == Subscriber.Input {
+        let subscriberIdentifier = subscriber.combineIdentifier
+
+        guard !subscriberIdentifiers.contains(subscriberIdentifier) else {
+            subscriber.receive(subscription: Subscriptions.empty)
+            return
+        }
+
+        let subscription = Subscription(downstream:
+        AnySubscriber(subscriber)) { [weak self] in
```
freak4pc, line 72: Sorry for the nitpick, I'll just commit this in
```suggestion
        let subscription = Subscription(downstream: AnySubscriber(subscriber)) { [weak self] in
```

Tests/ReplaySubjectTests.swift
```diff
@@ -0,0 +1,342 @@
...
+
+    func testReplaysNoValues() {
+        let subject = ReplaySubject<Int, Never>(bufferSize: 1)
+
+        var results = [Int]()
+
+        cancellables = Set([
+            subject
+                .sink(receiveValue: { results.append($0) })
+            ])
```
freak4pc, line 27: Why not `.store(in: cancellables)` ? Would go this route just to use the "standard" way and not confuse anyone with the syntax

jasdev: Ah word, yea I can swap that in!

jasdev: (incoming in the next commit)

PR #25, opened 2020-04-18
Adds conveniences to `AnyPublisher`

Prince2k3:
I've added some conveniences to `AnyPublisher` (with tests).  I thought it could be useful. If you guys do like them. I would like some ideas of how to add this to the ReadMe

Thanks!

@freak4pc


Sources/Extensions/AnyPublisher+Helpers.swift
```diff
@@ -0,0 +1,108 @@
...
+    static func just(_ output: Output) -> AnyPublisher<Output, Never> {
+        return Just(output).eraseToAnyPublisher()
+    }
+
+    /**
+    Returns a non-terminating observable sequence, which can be used to denote an infinite duration.
+    - returns: A publisher whose observers will never get called.
+    */
+    static func never() -> AnyPublisher<Never, Failure> {
+        return Never.NeverPublisher().eraseToAnyPublisher()
```
freak4pc, line 53:
```suggestion
        return Empty(completeImmediately: false).eraseToAnyPublisher()
```

freak4pc:
Hey! First of all, thank you so much for your contribution, it's not obvious to take from your time to write some code and send a PR :) 

I have a few minor issues with this PR, that doesn't mean we can't pull off some, or all, of these methods, but that it requires more discussion and thinking if it's _worth_ adding the extra code for things that are relatively easy to add if a consumer wants them: 

1. As is this PR seems too large to me, scope-wise. You've add _a lot_ of helpers, so it makes it a tad harder to discuss each one individually. 
2. Since this only applies to `AnyPublisher`, it has a very narrow use-case scope IMO, that highly depends on a team's/developer's programming methodology (e.g. if they erase everywhere or not) 
3. Lastly, there are many cases where the helpers end up being longer than the full syntax, which seems counter-intuitive to me. It also means we have places where the uses of Just and Fail are mehods instead of Apple's definition of having them as concrete publisher types (the method thing doesn't bother me as much but mentioning just in case):

```diff
- AnyPublisher<String, Never>.just("Hello")
+ Just("Hello").eraseToAnyPublisher()
```

```diff
- AnyPublisher<String, MyError>.fail(MyError.failure)
+ Fail(MyError.failure).eraseToAnyPublisher()
```

etc... 

It seems to me these would mostly be beneficial in cases of available type-inference, so it makes their scope even narrower. 

As is, I'd need to see a significant demand for adding these helpers to pull them in, because they don't seem generic enough or have a wide enough use-case scope to add to Ext. This is just my $0.02 obviously, and hope it doesn't prevent you from adding more PRs later on because the code itself and documentation are very well done! 

Thank you :)

Prince2k3:
@freak4pc 

No worries. This won't stop my contributions. I think the CombineExt library is a great addition to Combine framework. These are some ideas I thought were useful. 

My feedback to your response is you are correct. The helpers do help in the cause of type inference. Without it I still would recommend what you said, which is fallback to the normal use.

PR #27, opened 2020-04-28
Adds Publisher.prefix(duration:tolerance:on:in:options:)

jasdev:
Also, scouted a couple of linter issues in WithLatestFrom.swift. If you want me to split that out into a separate PR, lemme know. 👌🏽

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ All operators, utilities and helpers respect Combine's publisher contract, inclu
 * [setOutputType(to:)](#setOutputType)
 * [removeAllDuplicates and removeAllDuplicates(by:) ](#removeAllDuplicates)
 * [share(replay:)](#sharereplay)
-* [prefix(duration:tolerance:on:in:options:)](#prefixdurationtoleranceoninoptions)
+* [prefix(duration:tolerance:on:in:options:)](#prefixduration)
 
 ### Publishers
 * [AnyPublisher.create](#AnypublisherCreate)
@@ -471,7 +471,7 @@ second subscriber: 3
 second subscriber: 4
 ```
 
-### prefix(duration:tolerance:on:in:options:)
+### prefix(duration:)
 
 A time-based overload on `Publisher.prefix` that accepts value and completions events from an upstream publisher for a specified `duration` of time (in seconds). After expiring, the operator will emit a `.finished` event and ignore all future output. Configuration options are also exposed for the underlying timer, if needed.
 
--- a/README.md
+++ b/README.md
@@ -473,7 +473,7 @@ second subscriber: 4
 
 ### prefix(duration:)
 
-A time-based overload on `Publisher.prefix` that accepts value and completions events from an upstream publisher for a specified `duration` of time (in seconds). After expiring, the operator will emit a `.finished` event and ignore all future output. Configuration options are also exposed for the underlying timer, if needed.
+An overload on `Publisher.prefix` that that republishes values for a provided `duration` (in seconds), and then completes.
 
 ```swift
 let subject = PassthroughSubject<Int, Never>()
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ All operators, utilities and helpers respect Combine's publisher contract, inclu
 * [setOutputType(to:)](#setOutputType)
 * [removeAllDuplicates and removeAllDuplicates(by:) ](#removeAllDuplicates)
 * [share(replay:)](#sharereplay)
+* [prefix(duration:tolerance:on:in:options:)](#prefixdurationtoleranceoninoptions)
 
 ### Publishers
 * [AnyPublisher.create](#AnypublisherCreate)
@@ -468,6 +469,32 @@ first subscriber: 4
 second subscriber: 2
 second subscriber: 3
 second subscriber: 4
+```
+
+### prefix(duration:tolerance:on:in:options:)
+
+A time-based overload on `Publisher.prefix` that accepts value and completions events from an upstream publisher for a specified `duration` of time (in seconds). After expiring, the operator will emit a `.finished` event and ignore all future output. Configuration options are also exposed for the underlying timer, if needed.
+
+```swift
+let subject = PassthroughSubject<Int, Never>()
+
+subscription = subject
+  .prefix(duration: 0.5)
+  .sink(receiveValue: { print($0) })
+  
+subject.send(1)
+subject.send(2)
+subject.send(3)
+
+DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
+  subject.send(4)
+}
+```
+
+```none
+1
+2
+3
 ```
 
 ## Publishers
--- a/Sources/Operators/PrefixDuration.swift
+++ b/Sources/Operators/PrefixDuration.swift
@@ -10,7 +10,7 @@ import Combine
 import Foundation
 
 public extension Publisher {
-    /// Prefixes `self` for a length of time specified by `duration` (in seconds). Any value or completion events are sent downstream prior to the `duration` deadline. After then, a `.finished` event is fired and all subsequent value or completion events from `self` are ignored.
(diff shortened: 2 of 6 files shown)


Sources/Operators/PrefixDuration.swift
```diff
@@ -0,0 +1,30 @@
...
+    ///   - runLoop: The run loop for the underlying timer.
+    ///   - mode: The run loop mode for the underlying timer.
+    ///   - options: The run loop scheduler options for the underlying timer.
+    ///
+    /// - returns: A publisher that prefixes for a `duration` length of time and then finishes normally, if it hasn’t already before then.
+    func prefix(duration: TimeInterval,
+                tolerance: TimeInterval? = nil,
+                on runLoop: RunLoop = .main,
+                in mode: RunLoop.Mode = .default,
+                options: RunLoop.SchedulerOptions? = nil) -> Publishers.PrefixUntilOutput<Self, Publishers.Autoconnect<Timer.TimerPublisher>> {
```
freak4pc, line 27: I'm contemplataing if we should erase the "AutConnect->TimerPublisher" portion. It seems like non-interesting information to expose, as opposed to PrefixUntilOutput which is more coree info. WDYT? 

```suggestion
                options: RunLoop.SchedulerOptions? = nil) -> Publishers.PrefixUntilOutput<Self, Publishers.Autoconnect<Timer.TimerPublisher>> {
```

Sources/Operators/PrefixDuration.swift
```diff
@@ -0,0 +1,30 @@
...
+    /// Prefixes `self` for a length of time specified by `duration` (in seconds). Any value or completion events are sent downstream prior to the `duration` deadline. After then, a `.finished` event is fired and all subsequent value or completion events from `self` are ignored.
+    ///
+    /// - parameters:
+    ///   - duration: The time interval during which to accept value or completion events (in seconds).
+    ///   - tolerance: The tolerance the underlying timer.
+    ///   - runLoop: The run loop for the underlying timer.
+    ///   - mode: The run loop mode for the underlying timer.
+    ///   - options: The run loop scheduler options for the underlying timer.
+    ///
+    /// - returns: A publisher that prefixes for a `duration` length of time and then finishes normally, if it hasn’t already before then.
```
freak4pc, line 22: Would align a bit with Apple's docs:

```suggestion.
    /// Republishes elements for a specified duration.
    ///
    /// - parameters:
    ///   - duration: The time interval during which to accept value or completion events (in seconds).
    ///   - tolerance: The tolerance the underlying timer.
    ///   - runLoop: The run loop for the underlying timer.
    ///   - mode: The run loop mode for the underlying timer.
    ///   - options: The run loop scheduler options for the underlying timer.
    ///
    /// - returns: A publisher that republishes up to the specified duration.
```

Sources/Operators/WithLatestFrom.swift
```diff
@@ -8,6 +8,8 @@
 
 import Combine
 
+// swiftlint:disable large_tuple
```
freak4pc, line 11: Where are we using a large tuple? Might be worth just relaxing that rule

Tests/PrefixDurationTests.swift
```diff
@@ -0,0 +1,123 @@
...
+            .prefix(duration: 0.5)
+            .sink(receiveCompletion: { _ in expectation.fulfill() },
+                  receiveValue: { results.append($0) })
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.25) {
+            subject.send(1)
+        }
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1.5) {
+            subject.send(2)
```
freak4pc, line 32: we have to have a a test for more than a single value (preferably 3-4 values)

Sources/Operators/WithLatestFrom.swift
jasdev: Without disabling it, we’re hitting three warnings in the file.
Screen Shot 2020-05-03 at 6 25 06 PM

jasdev:
Should be good for another pass over, thanks for reviewing it! 🙏🏽

README.md
```diff
@@ -470,6 +471,32 @@ second subscriber: 3
 second subscriber: 4
 ```
 
+### prefix(duration:)
+
+A time-based overload on `Publisher.prefix` that accepts value and completions events from an upstream publisher for a specified `duration` of time (in seconds). After expiring, the operator will emit a `.finished` event and ignore all future output. Configuration options are also exposed for the underlying timer, if needed.
```
freak4pc, line 476: This explanation is a bit verbose and has a lot of impementation details, I think it would be better to just mention it's an overload of `prefix` that republishes values for the provided duration, and then completes

Sources/Operators/PrefixDuration.swift
```diff
@@ -0,0 +1,32 @@
...
+    ///
+    /// - returns: A publisher that republishes up to the specified duration.
+    func prefix(duration: TimeInterval,
+                tolerance: TimeInterval? = nil,
+                on runLoop: RunLoop = .main,
+                in mode: RunLoop.Mode = .default,
+                options: RunLoop.SchedulerOptions? = nil) -> Publishers.PrefixUntilOutput<Self, AnyPublisher<Date, Never>> {
+        prefix(untilOutputFrom: Timer.publish(every: duration, tolerance: tolerance, on: runLoop, in: mode, options: options)
+            .autoconnect()
+            .eraseToAnyPublisher())
```
freak4pc, line 30: Final nit here, I think the date isn't too interesting. WDYT of just going

```suggestion
                options: RunLoop.SchedulerOptions? = nil) -> Publishers.PrefixUntilOutput<Self, AnyPublisher<Void, Never>> {
        prefix(untilOutputFrom: Timer.publish(every: duration, tolerance: tolerance, on: runLoop, in: mode, options: options)
            .autoconnect()
            .map { _ in }
            .eraseToAnyPublisher())
```

Tests/PrefixDurationTests.swift
```diff
@@ -0,0 +1,151 @@
...
+        var completions = [Subscribers.Completion<Never>]()
+
+        cancellable = subject
+            .prefix(duration: 0.5)
+            .sink(receiveCompletion: { completions.append($0); expectation.fulfill() },
+                  receiveValue: { results.append($0) })
+
+        subject.send(1)
+        subject.send(2)
+        subject.send(3)
```
freak4pc, line 138: Can we have one of these delayed within the window? Also maybe move the test up after the single event test, so it's a bit chronological

Sources/Operators/PrefixDuration.swift
```diff
@@ -0,0 +1,32 @@
...
+    ///   - mode: The run loop mode for the underlying timer.
+    ///   - options: The run loop scheduler options for the underlying timer.
+    ///
+    /// - returns: A publisher that republishes up to the specified duration.
+    func prefix(duration: TimeInterval,
+                tolerance: TimeInterval? = nil,
+                on runLoop: RunLoop = .main,
+                in mode: RunLoop.Mode = .default,
+                options: RunLoop.SchedulerOptions? = nil) -> Publishers.PrefixUntilOutput<Self, AnyPublisher<Date, Never>> {
+        prefix(untilOutputFrom: Timer.publish(every: duration, tolerance: tolerance, on: runLoop, in: mode, options: options)
```
jdisho, line 28: Wondering, why is Apple using`RunLoop` and not `DispatchQueue`?

jasdev:
> What can be a use case for this operator?

A use case that came to mind for me is sampling from a publisher with a high value count—e.g. a sequence backing the Twitter firehose and you want to peak into it for a few seconds.

Sources/Operators/PrefixDuration.swift
jasdev: Not sure! Maybe some historical reason. 🤔

ohayon:
\o/ !!

PR #28, opened 2020-05-01
Fix double-subscribe bug in ReplaySubject.

jasdev:
👋🏽

I found a small inconsistency with how `ReplaySubject` handles double subscribes compared to `CurrentValue-` and `PassthroughSubject` (cc’ing @tcldr, so this is also on Entwine’s radar).

It turns out my initial assumption that double subscribes were ignored wasn’t actually inherent to `Subject`s and is instead a nuance of `Sink`. The folks behind OpenCombine [noticed this](https://github.com/broadwaylamb/OpenCombine/blob/d680f09932fe68942c3d391d6df296d38b9dcd05/Sources/OpenCombine/Subscribers/Subscribers.Sink.swift#L27), too. `Sink`s keep an internal subscription state that prevents subscriptions from doubly coming through.

I dig into this in greater detail [here](https://jasdev.me/publisher-temperature-primer).

We can verify that `Sink` keeps state by instead double subscribing to a `PassthroughSubject` with a more barebones `AnySubscriber` and seeing events logged twice.

```swift
import Combine

let subject = PassthroughSubject<Int, Never>()

let subscriber = AnySubscriber<Int, Never>(
	receiveSubscription: { $0.request(.unlimited) },
	receiveValue: { print($0); return .none },
	receiveCompletion: { print($0) }
)

subject
	.subscribe(subscriber)

subject
	.subscribe(subscriber)

subject.send(1)
subject.send(2)
subject.send(completion: .finished)
```

Logs:

```none
1
1
2
2
finished
finished
```

This PR updates `ReplaySubject` to follow suit and adjusts the related test, accordingly.

### DIFF
--- a/Tests/ReplaySubjectTests.swift
+++ b/Tests/ReplaySubjectTests.swift
@@ -286,30 +286,6 @@ final class ReplaySubjectTests: XCTestCase {
         XCTAssertTrue(completed)
     }
 
-    func testDoubleSubscribe() {
-        let subject = ReplaySubject<Int, Never>(bufferSize: 1)
-
-        subject.send(1)
-        subject.send(completion: .finished)
-
-        var results = [Int]()
-        var completions = [Subscribers.Completion<Never>]()
-
-        let sink1 = Subscribers.Sink<Int, Never>(
-            receiveCompletion: { completions.append($0) },
-            receiveValue: { results.append($0) }
-        )
-
-        subject
-            .subscribe(sink1)
-
-        subject
-            .subscribe(sink1)
-
-        XCTAssertEqual(results, [1])
-        XCTAssertEqual(completions, [.finished])
-    }
-
     private var demandSubscription: Subscription!
     func testRespectsDemand() {
         let subject = ReplaySubject<Int, Never>(bufferSize: 4)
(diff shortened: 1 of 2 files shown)


Tests/ReplaySubjectTests.swift
```diff
@@ -341,4 +317,30 @@ final class ReplaySubjectTests: XCTestCase {
...
+        )
+
+        subject
+            .subscribe(subscriber)
+
+        subject
+            .subscribe(subscriber)
+
+        XCTAssertEqual([2, 2], results)
+        XCTAssertEqual([.finished, .finished], completions)
```
freak4pc, line 344: Perfect! The only change I'd make is give a salt here so you can actually identify these are two different subjects and not two emissions of the same subject, i.e.:


```suggestion
        subject
            .map { "a\($0)")
            .subscribe(subscriber)

        subject
            .map { "b\($0)") 
            .subscribe(subscriber)

        XCTAssertEqual(["a2", "b2"], results)
        XCTAssertEqual([.finished, .finished], completions)
```

Tests/ReplaySubjectTests.swift
```diff
@@ -341,4 +317,32 @@ final class ReplaySubjectTests: XCTestCase {
...
     }
+
+    func testDoubleSubscribe() {
+        let subject = ReplaySubject<Int, Never>(bufferSize: 1)
+
+        subject.send(1)
+        subject.send(2)
+        subject.send(completion: .finished)
+
+        var results = [Int]()
```
freak4pc, line 328:
```suggestion
        var results = [String]()
```

Tests/ReplaySubjectTests.swift
```diff
@@ -341,4 +317,32 @@ final class ReplaySubjectTests: XCTestCase {
...
+        var completions = [Subscribers.Completion<Never>]()
+
+        let subscriber = AnySubscriber<Int, Never>(
```
freak4pc, line 331:
```suggestion
        let subscriber = AnySubscriber<String, Never>(
```

jasdev:
Done and done, should be good for another pass over. 🙏🏽

freak4pc:
That was probably super hard to track down. Impressive work! Thanks 🥇

jasdev:
Kudos to @broadwaylamb and OpenCombine’s contributors—I only caught this after a lot of head scratching and it was incredible to see that they already accounted for these details in their implementation. 💯

tcldr:
Ah ha, makes sense – I must admit I’d made a bit of an assumption in my own implementation. Great find! Agree the work that OpenCombine have done is seriously impressive. I wonder if WWDC will bring Apple’s own implementation of ReplaySubject?

jasdev:
> I wonder if WWDC will bring Apple’s own implementation of `ReplaySubject`?

I hope they do! Without it, folks are going to quickly realize they can’t share with replaying. 😅

PR #29, opened 2020-05-03
Lower deployment target and add availability checks to all Combine related code

RonKliffer:


freak4pc:
Me and @RonKliffer had a discussion around this yesterday, after I've heard this use-case from several people.

There are people who actually use Combine in newer portions of their apps, e.g. iOS 11 apps that want to use this dependency with an availability clause. I think it's a valid use-case overall. WDYT @jasdev @jdisho ?

Sources/Operators/CombineLatestMany.swift
```diff
@@ -8,6 +8,7 @@
 
 import Combine
 
+@available(iOS 13.0, *)
```
freak4pc, line 11: All tests should stay iOS 13 and up. Only the target itself should be iOS 11+ 
I'm actually not sure it's possible, but worth checking

jdisho:
Seems valid case to me too.👌

jdisho:
Should we do the same thing for SPM?

jasdev:
> WDYT @jasdev @jdisho?

Cool with me—no concerns on my end as long as you also think this is the right call for 11.x+ folks. 👌🏽

PR #30, opened 2020-05-08
Added `assign(to: on: ownership:)` operator

dsk1306:
Using `assign(to:)` might produce a memory leak because it creates a strong reference to `object` (there is a great topic on [swift forums](https://forums.swift.org/t/does-assign-to-produce-memory-leaks/29546/2) about this). I think it might be useful to have `assignWeakly(to:)` operator as part of CombineExt.

### DIFF
--- a/Tests/AssignOwnershipTests.swift
+++ b/Tests/AssignOwnershipTests.swift
@@ -1,0 +1,86 @@
+//
+//  AssignOwnershipTests.swift
+//  CombineExt
+//
+//  Created by Dmitry Kuznetsov on 08/05/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import XCTest
+import Combine
+import CombineExt
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+class AssignOwnershipTests: XCTestCase {
+    var subscription: AnyCancellable!
+    var value = 0
+    var subject: PassthroughSubject<Int, Never>!
+
+    override func setUp() {
+        super.setUp()
+
+        subscription = nil
+        subject = PassthroughSubject<Int, Never>()
+        value = 0
+    }
+
+    override func tearDown() {
+        super.tearDown()
+
+        subject = nil
+    }
+
+    func testWeakAssign() {
+        checkSingleAssign(for: .weak)
+    }
+
+    func testStrongAssign() {
+        checkSingleAssign(for: .strong)
+    }
+
+    func testUnownedAssign() {
+        checkSingleAssign(for: .unowned)
+    }
+
+    func testWeakRetain() {
+        checkRetainValue(for: .weak) { initialRetainCount, resultRetainCount in
+            XCTAssertEqual(initialRetainCount, resultRetainCount)
+        }
+    }
+
+    func testStrongRetain() {
+        checkRetainValue(for: .strong) { initialRetainCount, resultRetainCount in
+            XCTAssertEqual(initialRetainCount + 1, resultRetainCount)
+        }
+    }
+
+    func testUnownedRetain() {
+        checkRetainValue(for: .unowned) { initialRetainCount, resultRetainCount in
+            XCTAssertEqual(initialRetainCount, resultRetainCount)
+        }
+    }
+
+    private func checkSingleAssign(for ownership: ObjectOwnership) {
+        subscription = subject
+            .assign(to: \.value, on: self, ownership: ownership)
+
+        let newValue1 = 10
+        subject.send(newValue1)
+        XCTAssertEqual(value, newValue1)
+
+        let newValue2 = 11
+        subject.send(newValue2)
+        XCTAssertEqual(value, newValue2)
+    }
+
+    private func checkRetainValue(for ownership: ObjectOwnership, completion: (Int, Int) -> Void) {
+        let initialRetainCount = CFGetRetainCount(self)
+
+        subscription = subject
+            .assign(to: \.value, on: self, ownership: ownership)
+
+        let resultRetainCount = CFGetRetainCount(self)
+
+        completion(initialRetainCount, resultRetainCount)
+    }
+}
--- a/Tests/AssignOwnershipTests.swift
+++ b/Tests/AssignOwnershipTests.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2020 Combine Community. All rights reserved.
 //
 
+#if !os(watchOS)
 import XCTest
 import Combine
 import CombineExt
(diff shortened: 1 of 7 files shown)


freak4pc:
This is great! Thank you so much. I think we had some discussions here around this, and there are two thoughts I had

1. I wonder it's worth doing something like:

```swift
public extension Publisher where Failure == Never {
    func assign<Root: AnyObject>(to keyPath: ReferenceWritableKeyPath<Root, Self.Output>,
                                 on object: Root,
                                 retainment: RetainmentStrategy) -> AnyCancellable {
        switch retainment {
        case .strong:
            return assign(to: keyPath, on: object)
        case .weak:
            return sink(receiveValue: { [weak object] value in
                object?[keyPath: keyPath] = value
            })
        }
    }
}

public enum RetainmentStrategy {
    case strong
    case weak
}
```

The reasoning for this is that we might need to have a lot of these `weakly`-suffixed operators, and I'm wondering if it's better making an overload with some specific arguments. Thoughts ? @jasdev @jdisho 

2. Perhaps we should add the same for all `assign` overloads that accept many objects (e.g. AssignToMany.swift in the project)

Would love your thoughts :)

dsk1306:
I think it's a great idea. I can start to work on it unless @jasdev or @jdisho have something against it.

jasdev:
No concern on my end! I’d love to see something of this form land (I [tried back when](https://github.com/CombineCommunity/CombineExt/pull/14) hah). There’s definitely a need for it in every day Combine code (especially `ObservableObject` conformances) and for what it’s worth, the CXExtensions [think so, too](https://github.com/cx-org/CXExtensions/blob/3a777360ef0ec92342a2ed82fe748a78efead181/Sources/CXExtensions/WeakAssign.swift#L4-L11).

dsk1306:
👍

jdisho:
Hey folks 👋
Couple of questions from my side, but I'm not objecting what @freak4pc suggested. 
Since strongly assigned references cause memory leaks, why are we taking it into consideration? 
Should't we always want to weakly assign? Is there any use case in which we would like to strongly assign? 

This should behave similarly with RxSwift's `DisposeBag`, right?

jdisho:
Seems good to me! 🎉
Make sure to update the README too.

Sources/Operators/AssignWeakly.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  AssignWeakly.swift
```
jdisho, line 2: No sure if this should be called `AssignWeakly`. It's not just about the weakly part.

freak4pc:
@jasdev I know you tried, I think the thing remains the same from my angle 

Does it make sense having a bunch of "new" operators called `assignWeakly`, `sinkWeakly`, whatever, or should we overload "existing" operators with an argument like I mentioned above.

I'm still divided in this, but leaning towards argument overloading. Do you, as consumers, have a preference around this? Perhaps I ask on twitter to gain some clearer understanding. 

I see there's some prior art in CXExtensions so that's also something worth considering.

Sources/Operators/AssignWeakly.swift
dsk1306: I've renamed it to just `Assign`

jdisho:
Try making a poll in Twitter 😊

jasdev:
> Does it make sense having a bunch of “new” operators called `assignWeakly`, `sinkWeakly`, whatever, or should we overload “existing” operators with an argument like I mentioned above.

Agree that it should be an overload. 👌🏽 For the “bunch of” bit, I’m not too worried about that because the only “subscribers as operators” in Combine are `sink`, `assign`, and `subscribe`.

- `sink` already requires closure arguments, so consumers can tune capture semantics.
- `subscribe` comes with hint that the argument is likely `strong`ly captured (will depend on the underlying implementation).
- `assign` is focusing in on a path (or the whole with `\.self`) and since there’s no closure arguments, it can’t handle that `weak` capture we’re looking for.

The only note on my end with the `RetainmentStrategy` is we’ll probably want give folks an `RetainmentStrategy.unowned` case. 😄

dsk1306:
I wonder how many people would use it 🤔 But I agree, `unowned` should be a part of `RetainmentStrategy`.

freak4pc:
Ok so after a relatively wide survey in one of the slack channels and some friends, let’s go with:

`assign(to:on:ownership:)`

ownership: 
  `strong` (alias to `assign(to:on:)` )
  `weak`
  `unowned`

@dsk1306 @jasdev @jdisho

freak4pc:
I squashed and cleaned up a bit @dsk1306. 
The only thing that I think is left is that the to-many-ownership tests don't seem to actually provide coverage, not sure why: 

image

Tests/AssignToManyTests.swift
```diff
@@ -64,6 +67,77 @@ class AssignToManyTests: XCTestCase {
...
+        let initialRetainCount1 = CFGetRetainCount(dest1)
+        let initialRetainCount2 = CFGetRetainCount(dest2)
+
+        subscription = source
+            .assign(to: \.prop, on: dest1, and: \.ivar, on: dest2, ownership: ownership)
+
+        let resultRetainCount1 = CFGetRetainCount(dest1)
+        let resultRetainCount2 = CFGetRetainCount(dest2)
+
+        completion([(initialRetainCount1, resultRetainCount1),
```
freak4pc, line 114: Your tests don't actually seem to check the falues of dest1 and dest2 changes, right? 
Also seems like you're not pushing any value into `source` which might be the reason it's not covered

freak4pc:
Ok, updated the missing tests. Let me know what you think and if you want to make any additional changes before pulling this in and cutting the next version @dsk1306.

dsk1306:
I think it would be nice to move all ownership tests from `AssignToManyTests` to `AssignOwnershipTests`. Give me a sec, I'll do it.

freak4pc:
@dsk1306 I'm a bit iffy about this because it fits better into AssignMany than AssignOwnership. There's a bit of a fine line there, so I'm not entirely sure. Don't feel too strongly about it, I guess we can go either way.

dsk1306:
@freak4pc I've pushed one more commit with updated unit tests. Feel free to ignore it if you think it's unnecessary.

Tests/AssignOwnershipTests.swift
```diff
@@ -0,0 +1,98 @@
...
+
+        subscription = nil
+        subject = PassthroughSubject<Int, Never>()
+        value1 = 0
+        value2 = 0
+        value3 = 0
+    }
+
+    func testWeakOwnership() {
+        let initialRetainCount = CFGetRetainCount(self)
```
freak4pc, line 33: oh cool, great idea moving these here

PR #32, opened 2020-05-11
Adds Optional.publisher as a convenience for Optional.Publisher.init.

jasdev:


README.md
```diff
@@ -628,6 +631,26 @@ subject.send(5)
 5
 ```
 
+## Convenience
+
+## Optional.publisher
+
+`Optional.publisher` is a method version of [`Optional.Publisher.init`](https://developer.apple.com/documentation/swift/optional/publisher/3343960-init). It puts the type on equal footing with [`Result.publisher`](https://developer.apple.com/documentation/swift/result/3344716-publisher).
```
freak4pc, line 638:
```suggestion
`Optional.publisher` is a property version of [`Optional.Publisher.init`](https://developer.apple.com/documentation/swift/optional/publisher/3343960-init). It puts the type on equal footing with [`Result.publisher`](https://developer.apple.com/documentation/swift/result/3344716-publisher) and [`Sequence.publisher`](link here...).
```

README.md
```diff
@@ -628,6 +631,26 @@ subject.send(5)
...
+  /* … */
+```
+
+instead of
+
+```swift
+Optional.Publisher(Optional(1))
+  .publisher
+  /* … */
+```
```
freak4pc, line 652: Maybe like this? 

```suggestion
let number: Int? = 1
number.publisher
  /* … */

instead of

let number: Int? = 1
Optional.Publisher(number)
  /* … */
```

Sources/Convenience/Optional.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Created by Jasdev Singh on 11/05/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import Combine
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+public extension Optional {
+    /// A method form of `Optional.Publisher.init`, bringing `Optional` on the same footing with `Result.publisher`.
```
freak4pc, line 13: I'd use the same description as the original initialiser, I don't think docs should necessarily expose implementation details but mainly "what does this do"
```suggestion
    /// A publisher to emit the optional value of a successful result, or fail with an error.
```

Tests/OptionalTests.swift
```diff
@@ -0,0 +1,42 @@
...
+        var results = [Int]()
+        var completion: Subscribers.Completion<Never>?
+
+        subscription = Optional<Int>.none
+            .publisher
+            .sink(receiveCompletion: { completion = $0 },
+                  receiveValue: { results.append($0) })
+
+        XCTAssertTrue(results.isEmpty)
+        XCTAssertEqual(.finished, completion)
```
freak4pc, line 40: the docs mention this could end with an error, not sure if this is related to nullabiilty or just something confusing about the docs. I imagine it's just gonna compete without anything if it's `nil`?

Tests/PrefixDurationTests.swift
```diff
@@ -47,7 +47,7 @@ final class PrefixDurationTests: XCTestCase {
         var completions = [Subscribers.Completion<Never>]()
 
         cancellable = subject
-            .prefix(duration: 0.5)
+            .prefix(duration: 0.8)
```
freak4pc, line 50: Sneaky you

Tests/OptionalTests.swift
jasdev: `Optional.Publisher.Failure` is `Never`, so it can’t error. 😄

https://developer.apple.com/documentation/swift/optional/publisher/failure

README.md
```diff
@@ -628,6 +631,26 @@ subject.send(5)
...
+
+`Optional.publisher` is a property version of [`Optional.Publisher.init`](https://developer.apple.com/documentation/swift/optional/publisher/3343960-init). It puts the type on equal footing with [`Result.publisher`](https://developer.apple.com/documentation/swift/result/3344716-publisher) and [`Sequence.publisher`](https://developer.apple.com/documentation/swift/sequence/3344717-publisher).
+
+```swift
+let number: Int? = 1
+number.publisher
+  /* … */
+```
+
+instead of
```
freak4pc, line 646:
```suggestion
Instead of:
```

Sources/Convenience/Optional.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  Optional.swift
+//  CombineExt
+//
+//  Created by Jasdev Singh on 11/05/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import Combine
```
freak4pc, line 9: All new code should be wrapped in a 

```swift
#if canImport(Combine)

#endif
```

I might need to document this but it's something we have to do to support installation on older versions

Tests/OptionalTests.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created by Jasdev Singh on 11/05/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+import Combine
+import CombineExt
+import XCTest
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
```
freak4pc, line 13: Tests will need to be wrapped in `#if !os(watchOS)` since CocoaPods has issues with it otherwise when deploying pods 🤷

Tests/OptionalTests.swift
freak4pc: So the docs here are wrong, funny: 
https://developer.apple.com/documentation/swift/optional/publisher/3343960-init

freak4pc:
Resolved and merging since releasing today :)

jasdev:
Bless, thank you for carrying this over the merge line! 🙏🏽

PR #36, opened 2020-05-19
Add CURRENT_PROJECT_VERSION to target

danielob:
Appstore will reject app submission because `CFBundleVersion` is not available in the `info.plist` of the most current release binary `1.2.0`, this because the project is missing `CURRENT_PROJECT_VERSION`.

### DIFF
--- a/CombineExt.xcodeproj/project.pbxproj
+++ b/CombineExt.xcodeproj/project.pbxproj
@@ -469,6 +469,7 @@
 		OBJ_19 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				CURRENT_PROJECT_VERSION = 1;
 				ENABLE_TESTABILITY = YES;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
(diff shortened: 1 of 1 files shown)


CombineExt.xcodeproj/project.pbxproj
```diff
@@ -497,6 +498,7 @@
 		OBJ_20 /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				CURRENT_PROJECT_VERSION = 1;
```
nikita-leonov, line 501: This will add version 1 into bundle. Does bundle actually should be versioned to be inline with whatever release it is? So for the current master 1.3.0?

freak4pc:
I’m not sure what other frameworks do IRT SPM to be honest, 
Since the project is usually generated, these flags would also need to be added later on

nikita-leonov:
> I’m not sure what other frameworks do IRT SPM to be honest,
> Since the project is usually generated, these flags would also need to be added later on

Without this setting 1.3.0 release binaries were unfortunately generated incorrectly, as they again missing Bundle Version.

CombineExt.xcodeproj/project.pbxproj
danielob: Are we talking about current version (1.2.0) or a future version?

danielob:
> I’m not sure what other frameworks do IRT SPM to be honest,
> Since the project is usually generated, these flags would also need to be added later on

I dunno either and I have not enough experience using SPM to know what to be done. But I assume the value could be read from an XCConfig file?

freak4pc:
Hey, I'm going to solve this in a separate PR. Thanks for the report!

freak4pc:
Fixed in 745aee1

PR #38, opened 2020-06-06
Add toggle operator

kitwtnb:
Add toggle operator for boolean publisher.


freak4pc:
Thanks for the contribution! 

Looks great.. the only issue I have with this sort of operator is that its basically longer than what it wraps. E.g. `.map(!)` or `.map { !$0 }` are basically shorter than just `.toggle()`

The reason its useful in the standard library is that it actually makes things shorter by doing an in-place mutation.

So, practically speaking it doesn’t necessarily add much but perhaps its worth having just to be aligned with stdlib as sort of a convenience? WDYT @jasdev @jdisho

jdisho:
Even though it’s longer than `.map(!)` or `.map { !$0 }`, I think it’s more convenient. 
IMO, `.toggle()` reads better than `.map(!)`.

Since `Bool.toogle()` is a `mutating func` and this one here is not related to mutations, how about renaming it to `inverted()`? WDYT?

freak4pc:
To be honest, the only reason I consider adding it is to be aligned with stdlib... so having another different name doesn’t make sense to me... I think we can just pull it in as `toggle()`, it seems to convey the same meaning closely enough to be useful.

Sources/Operators/Toggle.swift
```diff
@@ -0,0 +1,21 @@
...
+#if canImport(Combine)
+import Combine
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+public extension Publisher where Output == Bool {
+    /// Toggle a boolean publisher.
+    ///
+    /// - returns: A toggled value.
+    func toggle() -> Publishers.Map<Self, Bool> {
+        map { !$0 }
```
freak4pc, line 18: `map(!)` could be shorter

Tests/ToggleTests.swift
```diff
@@ -0,0 +1,25 @@
...
+#if !os(watchOS)
+import Combine
+import CombineExt
+import XCTest
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+final class ToggleTests: XCTestCase {
+    func testSomeInitialization() {
+        var results = [Bool]()
+        _ = [true, false].publisher
```
freak4pc, line 18: Let’s add more values here

kitwtnb:
Thanks for review @freak4pc and @jdisho !

I fixed it, please check.

PR #40, opened 2020-06-07
Publisher.weakAssign(), like .assign but it will not retain

agiokas:
Publisher.weakAssign(), like .assign but it will not retain the object.

Assigning a long lived Publisher to a self.property using .assign it will create a reference to self and the object will not be able to deallocated until the subscription is canceled.

.weakassign tackles the issue. It will hold a weak reference to the object.


freak4pc:
Hey, thanks for the contribution but this already exists - see `assign(to:on:ownership:)`

PR #43, opened 2020-06-16
Remove expectations from the test suite in favor of TestSchedulers.

jasdev:
This PR pays down some test debt we’ve been sitting on without a `TestScheduler` type — and thankfully, the Point-Free folks open-sourced [theirs](https://github.com/pointfreeco/combine-schedulers)! The test suite went from running in roughly ~ten seconds locally (on a maxed-out ’19 MacBook Air) to under half a second.

Screen Shot 2020-06-15 at 9 43 11 PM


jasdev:
hmm — actually this would require us to bump CombineExt’s deployment targets considerably (from looking at the CI failures).

jasdev:
(closing for now until we find a way forward on the version targeting front — @mbrandonw mentioned he might be up for doing a similar `@availability` fencing in the schedulers repo.)

mbrandonw:
We're def down to having it added, but our schedules are a bit tight for the next week or more. If you've got time to PR it we'll happily take it, or we can sync up again sometime soon!

freak4pc:
I've made a PR over here: https://github.com/pointfreeco/combine-schedulers/pull/3

I switched Package.swift to my local branch and everything seems to run:
```swift
.package(url: "https://github.com/freak4pc/combine-schedulers", .branch("relax-deployment-target")),
```

jasdev:
(once CI is clear, should be good for a pass over @freak4pc — thank you for taking care of the versioning snag!)

Sources/Operators/PrefixDuration.swift
```diff
@@ -7,30 +7,39 @@
 //
 
 #if canImport(Combine)
+#if canImport(CombineSchedulers)
```
freak4pc, line 10: dont think this line is needed

Tests/CreateTests.swift
```diff
@@ -10,6 +10,7 @@
 import XCTest
 import Combine
 import CombineExt
+import CombineSchedulers
```
freak4pc, line 13: do you even use the test scheduler here or did you just remove expecations?

Tests/FlatMapLatestTests.swift
```diff
@@ -9,7 +9,8 @@
 #if !os(watchOS)
 import XCTest
 import Combine
-import CombineExt
+@testable import CombineExt
```
freak4pc, line 12: Why `@testable` is needed ?

jasdev: Great catch, this is a leftover change left from an intermediary commit that I later squashed.

Tests/CreateTests.swift
jasdev: Another lingering import, thanks for scouting it!

PR #50, opened 2020-08-19
Add ScanInto and ReduceInto

### LINKED ISSUE #13: ⚗️ Introducing `reduce(into:)`
Hi folks 👋

Since Swift Stdlib supports [`reduce(_:_:)`](https://developer.apple.com/documentation/swift/array/2298686-reduce#declarations) and [`reduce(into:_:)`](https://developer.apple.com/documentation/swift/array/3126956-reduce), how about introducing the Combine version of `reduce(into:_:)` into this library? 

- Just for convenience, it can be a wrapper around `reduce(_:_:)`, like the following: 

```swift 
public extension Publisher {
    func reduce<Result>(into initialResult: Result, _ nextPartialResult: @escaping (inout Result, Output) -> Void) -> Publishers.Reduce<Self, Result> {
        reduce(initialResult) { initial, nextPartial -> Result in
            var result = initial
            nextPartialResult(&result, nextPartial)
            return result
        }
    }
}
```

- Another option would be to create a new operator, e.g: `Publishers.ReduceInto<Upstream, Output>`. 

Which one do you think makes more sense? 😊

Personally, since the wrapper does not mimic `reduce(into:_:)` implementation from Stdlib, because it is just a wrapper around `reduce(_:_:)`, I'm leaning towards creating a new operator.

joewalsh:
Fixes #13 by introducing `scan(into:)`, `tryScan(into:)`,  `reduce(into:)`, and `tryReduce(into:)` extensions on `Publisher`.


freak4pc:
Uhm I don't remember exactly why we closed this :O 
Any chance you can rebase and clean this up so we can merge it? @joewalsh

freak4pc:
Ahhh because the base branch was changed to main. Can you just open a new PR possibly?

joewalsh:
> Ahhh because the base branch was changed to main. Can you just open a new PR possibly?
Will do!

Issue #37, opened 2020-05-27
Issues with Relays and .subscribe()


dsk1306:
I've faced an issue with Relays and `subscribe()` method when values aren't getting sent to the subscribed publisher.

I've created a simple VC to see what's going on.
```swift
class ViewController: UIViewController {
    let input = PassthroughSubject<Int, Never>()
    let output = PassthroughSubject<Int, Never>()
    var subsc = Set<AnyCancellable>()

    override func viewDidLoad() {
        super.viewDidLoad()

        input
            .subscribe(output)
            .store(in: &subsc)
        output
            .sink { print("Output", $0) }
            .store(in: &subsc)

        let fiveSecondsFromNow = DispatchTime.now() + 5
        DispatchQueue.main.asyncAfter(deadline: fiveSecondsFromNow) { [weak self] in
            self?.input.send(10)
        }
    }

}

```
I can see `Output 10` in the console output 5 seconds later, as expected.

However, if I replace `PassthroughSubject` with `PassthroughRelay`

```swift
class ViewController: UIViewController {
    let input = PassthroughRelay<Int>()
    let output = PassthroughRelay<Int>()
    var subsc = Set<AnyCancellable>()

    override func viewDidLoad() {
        super.viewDidLoad()

        input
            .subscribe(output)
            .store(in: &subsc)
        output
            .sink { print("Output", $0) }
            .store(in: &subsc)

        let fiveSecondsFromNow = DispatchTime.now() + 5
        DispatchQueue.main.asyncAfter(deadline: fiveSecondsFromNow) { [weak self] in
            self?.input.accept(10)
        }
    }
}
```

the output will be empty unless I replace `input.subscribe(output)` with `input.sink { [weak self] in self?.output.accept($0) }`.

Is it a bug or relays `subscribe` method doesn't suppose to work the same way as the one in regular subjects?

PR #54, opened 2020-09-02
Fixed issue with CurrentValueRelay and PassthroughRelay subscription

dsk1306:
Fix for https://github.com/CombineCommunity/CombineExt/issues/37

I've added two new unit tests to confirm this issue. It seems that when you try to subscribe one relay to another
```swift
let input = CurrentValueRelay<String>("initial")
let output = CurrentValueRelay<String>("initial")

input
    .subscribe(output)
    .store(in: &subscriptions)
output
    .sink(receiveCompletion: { _ in completed = true }, receiveValue: { self.values.append($0) })
    .store(in: &subscriptions)

input.accept("1")
input.accept("2")
input.accept("3")

XCTAssertFalse(completed)
XCTAssertEqual(values, ["initial", "1", "2", "3"])
```
`subscribe()` doesn't work because the test fails. You can confirm it by replacing `subscribe()` with `sink { output.accept($0) }`
```swift
let input = CurrentValueRelay<String>("initial")
let output = CurrentValueRelay<String>("initial")

input
    .sink { output.accept($0) }
    .store(in: &subscriptions)
output
    .sink(receiveCompletion: { _ in completed = true }, receiveValue: { self.values.append($0) })
    .store(in: &subscriptions)

input.accept("1")
input.accept("2")
input.accept("3")

XCTAssertFalse(completed)
XCTAssertEqual(values, ["initial", "1", "2", "3"])
```

### DIFF
--- a/Sources/Relays/CurrentValueRelay.swift
+++ b/Sources/Relays/CurrentValueRelay.swift
@@ -46,6 +46,10 @@ public class CurrentValueRelay<Output>: Relay {
         publisher.subscribe(storage)
     }
 
+    public func subscribe<P: Relay>(_ publisher: P) -> AnyCancellable where Output == P.Output {
+        publisher.subscribe(storage)
+    }
+
     deinit {
         // Send a finished event upon dealloation
         subscriptions.forEach { $0.forceFinish() }
--- a/Sources/Relays/PassthroughRelay.swift
+++ b/Sources/Relays/PassthroughRelay.swift
@@ -45,6 +45,10 @@ public class PassthroughRelay<Output>: Relay {
         publisher.subscribe(storage)
     }
 
+    public func subscribe<P: Relay>(_ publisher: P) -> AnyCancellable where Output == P.Output {
+        publisher.subscribe(storage)
+    }
+
     deinit {
         // Send a finished event upon dealloation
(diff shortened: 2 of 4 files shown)


dsk1306:
@freak4pc

Sources/Relays/CurrentValueRelay.swift
```diff
@@ -46,6 +46,10 @@ public class CurrentValueRelay<Output>: Relay {
         publisher.subscribe(storage)
     }
 
+    public func subscribe<P: Relay>(_ publisher: P) -> AnyCancellable where Output == P.Output {
+        publisher.subscribe(storage)
```
freak4pc, line 50: Thanks! I'm wondering what's the difference between this and `subscribe(_ publisher:)` ? It seems to be basically identical. Why does this solve the issue?

dsk1306: @freak4pc Honestly, I'm not quite sure because this is something I found experimentally.
It might be somehow connected to the fact that Relays do not conform to the Subject protocol, but I haven't found anything that could confirm it yet.

freak4pc: Alright, I think I'm OK with aligning it this way. Thanks!

PR #61, opened 2020-10-07
Added: StartWith, mapTo, and unwrap operators

SeRG1k17:
Perhaps these operators will be useful. If so, I can add tests and documentation later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ xcuserdata
 DerivedData/
 Carthage
 CombineExt.framework.zip
+*.DS_Store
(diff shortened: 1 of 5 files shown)


freak4pc:
Hey, I appreciate the good thought but I don't think these operators fit here:

- `unwrap` is simply `compactMap { $0 }` - we deprecated unwrap in RxSwiftExt as well since it's not needed anymore
- `startWith` is simply `prefix`
- `map(to: val)` is almost as long as `map { _ in val }`

I'm happy to accept other ideas you may have, but I'm not sure these specific options fit here. 

Thanks again!

SeRG1k17:
@freak4pc Thanks for the clarifications. I'm still in the process of switching from Rxswift to Combine, so there is confusion.

I don't quite agree with it `prefix`, it doesn't work like that because it limits the number of elements emitted like `take(:)` in Rx.

```swift
let numbers = (0...10)
cancellable = numbers.publisher
    .prefix(2)
    .sink { print("\($0)", terminator: " ") }

// Prints: "0 1"
```
```swift
let numbers = (0...10)
cancellable = numbers.publisher
    .startWith(2)
    .sink { print("\($0)", terminator: " ") }

// Prints: "2 0 1 2 .... 10"
```

freak4pc:
Sorry, I mean `prepend`. 

This might help:
https://github.com/CombineCommunity/rxswift-to-combine-cheatsheet

SeRG1k17:
Oh, yes, you right. Thank you for the link. If in the process of project development there are interesting ideas, I will try to place them here. Feel free to close it

freak4pc:
Thank you! Please keep sharing your ideas, those are always welcome and helpful :)

PR #62, opened 2020-10-17
Adds ignoreFailure(completeImmediately:) and ignoreFailure(setFailureType:completeImmediately:).

jasdev:
Mirroring `ignoreOutput` and `ignoreOutput(setOutputType:)` for the `Failure` generic. h/t @mbrandonw @stephencelis for [their discussion on this](https://github.com/pointfreeco/swift-composable-architecture/issues/271#issuecomment-682599562) and the feedback they filed (FB8563288).


freak4pc (approved):
It's a bit trivial but I guess I can see the motive for something like this. Thanks for the work and sorry for taking so long to look at this!

PR #68, opened 2020-11-22
Adds ConcatMap operator

### LINKED ISSUE #41: Add concatMap
As CombineExt already implements flatMapLatest, I think it might be a good idea to add concatMap as well. 

What do you think? I might look into it on a weekend if we agree on adding it. 😊

ohitsdaniel:
Resolves #41.

@freak4pc mentioned in #41, that we could try using `.flatMap(maxPublishers: 1, transform:)` to achieve the functionality of ConcatMap. I added some tests to verify if this is true, but unfortunately it's not that easy. ;) 

**Unexpected behaviour:**
* `.flatMap(maxPublishers: 1, transform:)` does not buffer mapped publishers, discarding all publishers if there is an active one, leading to missing values.

### DIFF
--- a/Sources/Operators/ConcatMap.swift
+++ b/Sources/Operators/ConcatMap.swift
@@ -8,7 +8,6 @@
 
 #if canImport(Combine)
 import Combine
-import class Foundation.NSRecursiveLock
 
 @available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
 public extension Publisher {
@@ -49,8 +48,8 @@ public extension Publishers {
             subscriber.receive(
                 subscription: Subscription(
                     upstream: upstream,
-                    transform: transform,
-                    downstream: subscriber
+                    downstream: subscriber,
+                    transform: transform
                 )
             )
         }
@@ -60,18 +59,21 @@ public extension Publishers {
 // MARK: - Subscription
 @available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
 private extension Publishers.ConcatMap {
-    final class Subscription<Downstream: Subscriber>: Combine.Subscription where Downstream.Input == Output, Downstream.Failure == Failure {
-        private var sink: Sink<Downstream>?
+    final class Subscription<Downstream: Subscriber>: Combine.Subscription where
+        Downstream.Input == NewPublisher.Output,
+        Downstream.Failure == Failure
+    {
+        private var sink: OuterSink<Downstream>?
 
         init(
             upstream: Upstream,
-            transform: @escaping Transform,
-            downstream: Downstream
+            downstream: Downstream,
+            transform: @escaping Transform
         ) {
-            self.sink = Sink(
+            self.sink = OuterSink(
                 upstream: upstream,
                 downstream: downstream,
-                transform: { transform($0) }
+                transform: transform
             )
         }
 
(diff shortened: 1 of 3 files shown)


ohitsdaniel:
A gave it a go and implemented Publishers.ConcatMap. As this is my first custom combine operator / Publisher, I'm happy for every feedback. 😸

Sources/Operators/ConcatMap.swift
```diff
@@ -0,0 +1,163 @@
...
+//  ConcatMap.swift
+//  CombineExt
+//
+//  Created by Daniel Peter on 22/11/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+#if canImport(Combine)
+import Combine
+import class Foundation.NSRecursiveLock
```
freak4pc, line 11: no reason for this, Combine already uses Foundation. I think import Foundation would be fine here.

Sources/Operators/ConcatMap.swift
```diff
@@ -0,0 +1,163 @@
...
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+public extension Publisher {
+    /// Transforms an output value into a new publisher, and flattens the stream of events from these multiple upstream publishers to appear as if they were coming from a single stream of events.
+    ///
+    /// Mapping to a new publisher will keep the subscription to the previous one alive until it completes and only then subscribe to the new one. This also means that all values sent by the new publisher are not forwarded as long as the previous one hasn't completed.
```
freak4pc, line 17:
```suggestion
    /// Mapping to a new publisher will keep the subscription to the previous one alive until it completes and only then subscribe to the new one. This also means that all values sent by the new publisher are not forwarded as long as the previous one hasn't completed.
```

> This also means that all values sent by the new publisher are not forwarded as long as the previous one hasn't completed.

Not sure about this - why would it emit at all if it's only subscribed to after the completion of the previous one?

Sources/Operators/ConcatMap.swift
```diff
@@ -0,0 +1,163 @@
...
+    ///
+    /// - parameter transform: A transform to apply to each emitted value, from which you can return a new Publisher
+    ///
+    /// - returns: A publisher emitting the values of all emitted publishers in order.
+    func concatMap<T, P>(
+        _ transform: @escaping (Self.Output) -> P
+    ) -> Publishers.ConcatMap<P, Self> where T == P.Output, P: Publisher, Self.Failure == P.Failure {
+        return Publishers.ConcatMap(upstream: self, transform: transform)
```
freak4pc, line 25:
```suggestion
        Publishers.ConcatMap(upstream: self, transform: transform)
```

Sources/Operators/ConcatMap.swift
```diff
@@ -0,0 +1,163 @@
...
+            lock.lock()
+            defer { lock.unlock() }
+
+            if activePublisher == nil {
+                setActivePublisher(mapped)
+            } else {
+                bufferedPublishers.append(mapped)
+            }
+
+            return .unlimited
```
freak4pc, line 124: Not 100% sure returning unlimited demand here is right. Any reason?

Sources/Operators/ConcatMap.swift
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        private func setActivePublisher(_ publisher: NewPublisher) {
+            lock.lock()
+            defer { lock.unlock() }
+            activePublisher = publisher
+
+            publisher.sink(
```
freak4pc, line 132: I'm not sure I'm a fan of using sink inside a custom publisher. I mean, you're alraedy holding your own subscription etc so why use a built-in subscriber that creates a subscription to do this ?

freak4pc (changes requested):
Still didn't go through the tests but have some questions. Thank you for all the hard work!

freak4pc:
Hey @ohitsdaniel - are you interested in finishing this PR? Thanks :)

ohitsdaniel:
Yes, just didn't find the time yet as I have been focusing on other projects. I'll see if I can squeeze it in next week. :)

ohitsdaniel:
Quick update from my side: I scheduled some time this Friday to work on this. :)

Sources/Operators/ConcatMap.swift
ohitsdaniel: The inner publisher could potentially emit before being subscribed to (Like a non-deferred future, i.e. hot observable)?


Example:
```swift
let passthrough = PassthroughSubject<Int, Never>()

let mapped = passthrough.map { value in
    Future<Int, Never> { completion in
        completion(.success(value))
    }
    .delay(for: .seconds(1), scheduler: DispatchQueue.main)
  }
  .compactMap { $0 }

  let cancellable = mapped.sink(receiveValue: { value in
    print(value)
  })

  passthrough.send(1)
  passthrough.send(2)     
```

In this case, concatMap does not **buffer** 2 as the Publisher 1 hadn't completed when 2 was put onto the stream.

See `test_ignores_values_of_subsequent_while_previous_hasNot_completed` in `CompactMapTests.swift`

Sources/Operators/ConcatMap.swift
ohitsdaniel: No reason, but I'm also not 100% sure what to return here. 🤔

Sources/Operators/ConcatMap.swift
ohitsdaniel: Problem is that we have an outer publisher of publishers and we want to flatten the value of the inner publisher into one publisher of values.

I checked how [RxSwift solves this](https://github.com/ReactiveX/RxSwift/blob/main/RxSwift/Observables/Merge.swift#L133) and it relies on `MergeLimitedSink` which subscribes a `MergeLimitedSinkIter` (inner sink) to each inner publisher and then checks if the outer sink [has a 'next' publisher](https://github.com/ReactiveX/RxSwift/blob/main/RxSwift/Observables/Merge.swift#L172) when the inner publisher completes.

The only benefit that I see from this is that it reuses 'active sinks' and subscribes them directly to the next publisher. Maybe you see some other benefits in this approach? I'm not 100% familiar with the RxSwift codebase.

ohitsdaniel: Just discovered that having an outer / inner sink would solve the whole uncertainty around which demand to return. I'll rewrite it to look a bit more like the RxSwift equivalent.

ohitsdaniel:
@freak4pc 

Re-implemented it, pretty closely following the RxSwift implementation while not over-generifying the implementation. Main idea is that we have an inner and outer sink.

The OuterSink observes the stream of upstream values and has .unlimited demand. Whenever it receives a value, it transforms it into a new publisher (NewPublisher) and hands it to the inner sink.

The InnerSink observes a stream of NewPublisher values and manages the demand. Whenever it receives a completion, it subscribes to the next publisher, if there is one. Whenever the outer sink hands in a NewPublisher, the inner sink either directly subscribes to the new publisher, if it currently has no active subscription, or queues it, so that it is subscribed to whenever the currently active subscription ends (i.e. the inner sink receives a successful completion).

Had to add a `remainingDemand` field to DemandBuffer, so that the inner sink can request more values if the downstream demand hasn't been fulfilled.

filblue:
@freak4pc  Any plans to move on with this?

Sources/Operators/ConcatMap.swift
danielt1263: In Rx, the concatMap assumes that inner Observables are cold and that the closure creates the Observable that is being subscribed to. I think it's correct to make that assumption here as well.

kerrmarin-lvmh:
Hey, are there any plans to merge this?

Evertt:
bump

freak4pc:

PR #70, opened 2020-12-30
Adds Publisher.prefix(while:behavior:).

jasdev:
Figured we should add this to Ext after stumbling upon @jessegrosjean’s [Swift Forums thread](https://forums.swift.org/t/how-to-make-combines-prefix-operator-inclusive/39216) and @mayoff’s [approach with composed operators](https://forums.swift.org/t/how-to-make-combines-prefix-operator-inclusive/39216/8).

### DIFF
--- a/Sources/Operators/PrefixWhileBehavior.swift
+++ b/Sources/Operators/PrefixWhileBehavior.swift
@@ -1,0 +1,78 @@
+//
+//  PrefixWhileBehavior.swift
+//  CombineExt
+//
+//  Created by Jasdev Singh on 29/12/2020.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+#if canImport(Combine)
+import Combine
+import Foundation
+
+/// Whether to include the first element that doesn’t pass
+/// the `while` predicate passed to `Combine.Publisher.prefix(while:behavior:)`.
+public enum PrefixWhileBehavior {
+    /// Include the first element that doesn’t pass the `while` predicate.
+    case inclusive
+
+    /// Exclude the first element that doesn’t pass the `while` predicate.
+    case exclusive
+}
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+public extension Publisher {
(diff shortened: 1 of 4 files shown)


Sources/Operators/PrefixWhileBehavior.swift
```diff
@@ -0,0 +1,77 @@
...
+        case .end:
+            return false
+        case .whileValueOrIncluded:
+            return true
+        }
+    }
+
+    var value: Output? {
+        switch self {
+        case .end:
```
freak4pc, line 71: this abstraction confuses me a bit - can you explain?

mayoff:
The goal is an operator that passes through all of the upstream values, until and **including** the first value that fails the predicate.

In the linked Swift forums thread, I implemented the requested operator as follows:

- Turn each upstream value into a publisher that emits the value (in a wrapper), followed by an end-marker if the value fails the predicate. In my original version, I used `Sequence.Publisher`. In @jasdev's code, the `Just ... append ... setFailureType` expression produces these short publishers.

- Use `flatMap` to flatten those short publishers into one publisher than republishes each upstream value, with an end-marker after each upstream value that fails the predicate.

- Use `prefix(while:)` to terminate the stream after the first end-marker.

- Use `compactMap` to unwrap the values and discard the end-marker.

In my original code, the wrappers and the end-marker were cases of `enum Cut`. Here, `@jasdev` has renamed `enum Cut` to `enum PrefixInclusiveEvent`.

Sources/Operators/PrefixWhileBehavior.swift
mayoff: See [my comment](https://github.com/CombineCommunity/CombineExt/pull/70#issuecomment-792062132) below.

jasdev:
(Cleaning up some ol’ PRs — @freak4pc any chance we could land this in the next release whenever you get a chance to review? 🙏🏽)

freak4pc (approved):
Sorry @jasdev & @mayoff,
Barely had any time lately and this totally slipped my mind. Thanks for the ping (and explanation), looks great.

PR #74, opened 2021-01-26
Fine tunes ReplaySubject’s locking.

jasdev:
The locking I added in https://github.com/CombineCommunity/CombineExt/pull/72 was a bit too board — @sharplet noted:

> [sending messages to subscribers] while holding a lock can lead to a recursive call trying to take the same lock again and crashing.

### DIFF
--- a/Sources/Subjects/ReplaySubject.swift
+++ b/Sources/Subjects/ReplaySubject.swift
@@ -23,7 +23,7 @@ public final class ReplaySubject<Output, Failure: Error>: Subject {
     private var buffer = [Output]()
 
     // Keeping track of all live subscriptions, so `send` events can be forwarded to them.
-    private var subscriptions = [Subscription<AnySubscriber<Output, Failure>>]()
+    private(set) var subscriptions = [Subscription<AnySubscriber<Output, Failure>>]()
 
     private var completion: Subscribers.Completion<Failure>?
     private var isActive: Bool { completion == nil }
--- a/Sources/Subjects/ReplaySubject.swift
+++ b/Sources/Subjects/ReplaySubject.swift
@@ -37,56 +37,75 @@ public final class ReplaySubject<Output, Failure: Error>: Subject {
     }
 
     public func send(_ value: Output) {
-        lock.lock()
-        defer { lock.unlock() }
+        let subscriptions: [Subscription<AnySubscriber<Output, Failure>>]
 
-        guard isActive else { return }
+        do {
+          lock.lock()
+          defer { lock.unlock() }
 
-        buffer.append(value)
+          guard isActive else { return }
 
-        if buffer.count > bufferSize {
+          buffer.append(value)
+          if buffer.count > bufferSize {
             buffer.removeFirst()
+          }
+
+          subscriptions = self.subscriptions
         }
 
         subscriptions.forEach { $0.forwardValueToBuffer(value) }
     }
 
     public func send(completion: Subscribers.Completion<Failure>) {
-        lock.lock()
-        defer { lock.unlock() }
+        let subscriptions: [Subscription<AnySubscriber<Output, Failure>>]
 
-        guard isActive else { return }
+        do {
+            lock.lock()
+            defer { lock.unlock() }
 
-        self.completion = completion
+            guard isActive else { return }
+
+            self.completion = completion
+
+            subscriptions = self.subscriptions
+        }
 
         subscriptions.forEach { $0.forwardCompletionToBuffer(completion) }
     }
 
     public func send(subscription: Combine.Subscription) {
-        lock.lock()
-        defer { lock.unlock() }
-
         subscription.request(.unlimited)
     }
 
     public func receive<Subscriber: Combine.Subscriber>(subscriber: Subscriber) where Failure == Subscriber.Failure, Output == Subscriber.Input {
-        lock.lock()
-        defer { lock.unlock() }
-        
         let subscriberIdentifier = subscriber.combineIdentifier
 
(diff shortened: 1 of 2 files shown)


Sources/Subjects/ReplaySubject.swift
```diff
@@ -37,53 +37,51 @@ public final class ReplaySubject<Output, Failure: Error>: Subject {
...
+        lock.unlock()
 
         if buffer.count > bufferSize {
+            lock.lock()
             buffer.removeFirst()
+            lock.unlock()
         }
 
         subscriptions.forEach { $0.forwardValueToBuffer(value) }
     }
```
sharplet, line 53: More fine-grained locking is moving in the right direction, but unfortunately this introduces some other problems:

- `isActive` is no longer protected by the lock, so accesses could race. This might result in thread 1 seeing `isActive == true` while thread 2 is trying to complete or cancel the subject.
- `buffer.append()` is locked, but then `buffer.count` is accessed immediately afterward. This could crash if thread 1 sees a count of 1, while thread 2 is removing the last item from the buffer, then calls `buffer.removeFirst()` on an empty buffer.
- Thread 2 could change `subscriptions` while thread 1 is sending the new value, resulting in an off-by-one error or subscribers missing events.

I'd recommend a pattern like this:

```swift
public func send(_ value: Output) {
  let subscriptions: [Subscription<AnySubscriber<Output, Failure>>]

  do { // critical section
    lock.lock()
    defer { lock.unlock() }

    guard isActive else { return }

    buffer.append(value)
    if buffer.count > bufferSize {
      buffer.removeFirst()
    }

    subscriptions = self.subscriptions
  }

  subscriptions.forEach { $0.forwardValueToBuffer(value) }
}
```

jasdev: Updated! Followed suit for the other methods and I thiiink I got it right — appreciate the review pass here. 🙏🏽

Sources/Subjects/ReplaySubject.swift
```diff
@@ -37,53 +37,74 @@ public final class ReplaySubject<Output, Failure: Error>: Subject {
...
+        let completion: Subscribers.Completion<Failure>?
 
-            self.subscriptions.remove(at: subscriptionIndex)
-        }
+        do {
+            lock.lock()
+            defer { lock.unlock() }
+
+            subscription = Subscription(downstream: AnySubscriber(subscriber)) { [weak self] in
+                guard let self = self else { return }
```
sharplet, line 92: I wonder if the body of this completion handler could be extracted to a separate method? I think having it nested here makes it a little difficult to reason about the two nested scopes where `lock` is referenced. They appear to overlap, but are actually separated in time.

Additionally, initializing the local `subscription` variable doesn't need to be protected by the lock, only the part where it's appended to `self.subscriptions`.

sharplet (commented):
One suggestion for restructuring the implementation of `receive(subscriber:)` to make the critical sections a little clearer, but otherwise looking good!

Sources/Subjects/ReplaySubject.swift
jasdev: > Additionally, initializing the local `subscription` variable doesn’t need to be protected by the lock, only the part where it’s appended to `self.subscriptions`.

Great catch. Fixing in the next commit. ✅

> I wonder if the body of this completion handler could be extracted to a separate method?

My only hesitation with refactoring as a method would be the implicit `self` capture — moving `subscription`’s initialization out of the `do` block should hopefully make the locking across time more clear.

sharplet: I wasn't thinking that capturing `self` would be implicit — something more along these lines:

```swift
let subscription = Subscription(downstream: AnySubscriber(subscriber)) { [weak self] in
  self?.completeSubscriber(withIdentifier: subscriberIdentifier)
}
```

jasdev: Ahh roger roger, I dig it! Will update in a few. 👌🏽

Sources/Subjects/ReplaySubject.swift
```diff
@@ -37,57 +37,79 @@ public final class ReplaySubject<Output, Failure: Error>: Subject {
...
-        self.completion = completion
+            subscriptions = self.subscriptions
+        }
 
         subscriptions.forEach { $0.forwardCompletionToBuffer(completion) }
     }
 
     public func send(subscription: Combine.Subscription) {
-        lock.lock()
-        defer { lock.unlock() }
```
sharplet, line 67: Good catch 👍

Sources/Subjects/ReplaySubject.swift
```diff
@@ -37,57 +37,79 @@ public final class ReplaySubject<Output, Failure: Error>: Subject {
...
     }
+
+    private func completeSubscriber(withIdentifier subscriberIdentifier: CombineIdentifier) {
+        lock.lock()
+        defer { self.lock.unlock() }
+
+        guard let subscriptionIndex = self.subscriptions
+                .firstIndex(where: { $0.innerSubscriberIdentifier == subscriberIdentifier }) else { return }
+
+        self.subscriptions.remove(at: subscriptionIndex)
```
sharplet, line 111: Now that this has been extracted, I'm wondering if it would make sense to refactor this to use `removeAll(where:)`?

jasdev: Even better — pushed that change up. 💯

freak4pc:
Thanks! Could this locking be relevant for other subjects as well? @jasdev

PR #79, opened 2021-03-05
Add mapToResult operator, tests

kaphacius:
This PR adds a new operator `catchToResult` which allows transforming a `Publisher` that has a specific `Failure` type (not `Never`) into a publisher that outputs `Output` and `Failure` wrapped in a `Result`. 
A common use-case for this is when an interface should produce publishers of type `AnyPublisher<Result<T, Error>, Never>`, but the underlying code works with publishers like `AnyPublisher<T, Error>` (`URLSession.DataTaskPublisher`).

The opposite situation, going from `AnyPublisher<Result<T, Error>, Never>` to `AnyPublisher<T, Error>` also happens. If there is an interest in this PR I can open another one with an operator that allows doing that.


Sources/Operators/CatchToResult.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+//  Created by Yurii Zadoianchuk on 05/03/2021.
+//  Copyright © 2021 Combine Community. All rights reserved.
+//
+
+#if canImport(Combine)
+import Combine
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+public extension Publisher where Failure: Error {
```
manmal, line 13: I think `Failure` always needs to conform to `Error`, so `where Failure: Error` could probably be removed?

Sources/Operators/CatchToResult.swift
```diff
@@ -0,0 +1,25 @@
...
+    /// Transform a publisher with concrete Output and Failure types
+    /// to a new publisher that wraps Output and Failure in Result,
+    /// and has Never for Failure type
+    /// - Returns: A type-erased publiser of type <Result<Output, Failure>, Never>
+    func catchToResult() -> AnyPublisher<Result<Output, Failure>, Never> {
+        self.map({ Result.success($0) })
+            .catch({ Just(.failure($0)) })
+            .eraseToAnyPublisher()
```
manmal, line 21: Nitpick: `self` isn't necessary here, and braces can be simplified:
```swift
map(Result.success)
    .catch { Just(.failure($0)) }
    .eraseToAnyPublisher()
```

manmal (commented):
I think this looks great, thanks for taking up the idea!

jasdev:
Would love to see this land, thanks @kaphacius! I use TCA’s [`catchToEffect` analog](https://github.com/pointfreeco/swift-composable-architecture/blob/e93527097282d39588d1d4c897b9323301785974/Sources/ComposableArchitecture/Effect.swift#L319-L335) often. ✅

> The opposite situation, going from `AnyPublisher<Result<Output, Failure>, Never>` to AnyPublisher<Output, Failure> also happens. If there is an interest in this PR I can open another one with an operator that allows doing that.

+1.

(and as a heads up, we’ll probs need some quick docs. in the README.md, for when you get a chance. 🙏🏽)

Tests/CatchToResultTests.swift
```diff
@@ -0,0 +1,153 @@
...
+
+    func testCatchCustomError() {
+        let subject = PassthroughSubject<Int, Error>()
+        var completed = false
+        var gotFailure = false
+        var gotSuccess = false
+        var result: Result<Int, Error>? = nil
+
+        subscription = subject
+            .tryMap({ _ -> Int in throw(CatchToResultError.someError) })
```
freak4pc, line 53:
```suggestion
            .tryMap { _ -> Int in throw CatchToResultError.someError }
```

please drop parentheses in trailing-closures everywhere

Also, throw isn't a function, it's a specialty word.

Tests/CatchToResultTests.swift
```diff
@@ -0,0 +1,153 @@
...
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+final class CatchToResultTests: XCTestCase {
+    private var subscription: AnyCancellable!
+
+    enum CatchToResultError: Error {
+        case someError
+    }
+
+    func testCatchToResult() {
```
freak4pc, line 24:
```suggestion
    func testCatchResultNoError() {
```

freak4pc (commented):
Thanks! Tests look good overall, but there are some formatting issues, for example usage of unneeded parentheses, using keywords like `catch` and `throw` as functions (e.g. `catch (let e)`, `throw(MyError.error)`. 

If you'll clean up these things I'll be happy to merge.

Thank you!

kaphacius:
Added info to README.md and applied all of the suggested comments 👍

jshier:
Is there any particular reason not to make the name of this something more general, like `toResult`? It's not strictly a catching operator, as it transforms the success case as well. 

(I have this operator in my code base as `asResult()`, though I'm not particular on `as` vs. `to`.)

freak4pc:
I agree about @jshier's remarks and this is the biggest gripe I have here. Something like maoToResult or something else makes more sense, but there's "prior art" with TCA's catchToEffect. Not sure that's really a strong enough reason to not give it our own naming.

kaphacius:
I don't have any attachment to the name to be honest, I think `mapToResult` sounds equally good. I'll make the changes and update the PR

kaphacius:
Renamed the operator, tests and the PR name :)

freak4pc (approved):
Looks great! Thank you :)

PR #82, opened 2021-03-12
Add liftErrorFromResult and tests

kaphacius:
This PR adds a new operator liftErrorFromResult which allows transforming a Publisher that outputs values wrapped in a `Result<T, E>` and `Failure` of `Never` to a new publisher `AnyPublisher<T,E>`.  
A common use-case for this is when the we expect to work with Non-`Never` publishers,  but the underlying code is producing publishers that have `Output` and `Failure` wrapped in a `Result` 

This is a continuation (and an inverse operator) of #79

### DIFF
--- a/README.md
+++ b/README.md
@@ -719,6 +719,37 @@ subscription = ints
 .finished
 ```
 
+### liftErrorFromResult
+
+Transforms a publisher of type `AnyPublisher<Result<T, E>, Never>` to `AnyPublisher<T, E>`
+
+```swift
+enum AnError: Error {
+    case someError
+}
+
+let subject = PassthroughSubject<Result<Int, AnError>, Never>()
+
+let subscription = subject
+    .liftErrorFromResult()
+    .sink(receiveCompletion: { print("completion: \($0)") },
+          receiveValue: { print("value: \($0)") })
+
+subject.send(.success(1))
+subject.send(.success(2))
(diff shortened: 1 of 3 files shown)


freak4pc (commented):
Thanks for the contribution! Next time it's better to open an issue to discuss first.

I'm not sure I 100% see the value of this operator:

1. It kinda provides the same value of `materialize()` and `dematerialize()`
2. It seems to cater for a very narrow use case 
3. It seems simple enough to add to the codebase of someone who needs it specifically for their use case.

Do other folks think this is a worthwhile addition? I'm a bit divided on it. 

Transforming a stream of values to stream of results seems reasonable but I'm not sure the other way around does. @jasdev @manmal

manmal:
I agree that this operator would probably not be used as often as its inverse. Personally, I've had something like this operator in one or two projects so far. But I think it's valuable to reject or postpone new features if they are not often used. I really don't have a strong opinion on this one.

jasdev:
No strong opinion — it’s nice for sake of completeness to offer operators for both directions, but if `catchToResult` is used more widely, let’s maybe start there and revisit if folks express a need for this. ✅

kaphacius:
Alright, since there is no strong support for this, I'll close it. Maybe let's revisit it in the future :)

PR #83, opened 2021-03-14
Make WithLatestFrom transparently CompactMap

b8591340:
It leverages `Sink.TransformOutput`‘s signature to merely “forward” the optional through `ResultSelector`.

Not sure if `WithLatestFrom` is supposed to drop `nil` values from `AnyPublisher<Optional<Int>, Never>`. If it is this merely compacts within the result selector as well. If not then it‘s a bug due to the way the sink‘s `transformOutput` returns the `resultSelector` and it needs to be addressed, and a compacting `WithLatestFrom`, if welcome, would likely need to be an operator by itself.

### DIFF
--- a/Sources/Operators/WithLatestFrom.swift
+++ b/Sources/Operators/WithLatestFrom.swift
@@ -21,9 +21,9 @@ public extension Publisher {
   ///
   ///  - returns: A publisher containing the result of combining each value of the self
   ///             with the latest value from the second publisher, if any, using the
-  ///             specified result selector function.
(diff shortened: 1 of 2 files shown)


freak4pc (commented):
Hey, thanks for this - but withLatestFrom isn't supposed to compactMap. 
If it does this for some reason, it's a mistake. What's the behavior you're seeing? Elements being dropped?

b8591340:
Well, it seems I might have had the failing test case in this branch.

I‘ll close, but if a flattening withLatestFrom is still welcome I‘ll look into implementing it myself.

Issue #85, opened 2021-03-29
Memory leak issue with ReplaySubject


jabeattie:
Hi Folks, we recently started using CombineExt in a project and it's already proven itself invaluable for fixing plenty of common scenarios we wanted to solve, so kudos for that.

We had a need to use `share(replay:)` recently and noticed after implementing it that we had some retain cycles appearing in our application. After spending some time looking into them, it _looks_ to be caused by `ReplaySubject` not releasing its subscriptions when they receive a completion.

I've got an idea for a fix, but I'm not well versed enough on Combine yet to appreciate what the other implications of it may be, so any help would be appreciated, thanks.

jabeattie:
Some screenshots from the memory graph after using the `share(replay:)` operator

<img width="1421" alt="Screenshot 2021-03-29 at 07 51 22" src="https://user-images.githubusercontent.com/6996767/112800166-34650900-9067-11eb-9c49-73b056715011.png">
<img width="856" alt="Screenshot 2021-03-29 at 07 51 36" src="https://user-images.githubusercontent.com/6996767/112800168-362ecc80-9067-11eb-827c-07e4c5c80361.png">

PR #86, opened 2021-03-29
Share replay leak

jabeattie:
Attempts to fix https://github.com/CombineCommunity/CombineExt/issues/85

I added a test that proved that the source publisher was retained after everything had been nilled out/released.

I then added the `cancel()` method call in the `forwardCompletionToBuffer` method inside ReplaySubject.Subscription as this causes the subscription to then be released.

As I mentioned in the issue I raised, I'm not familiar enough to know what the other implications of this could be (it seems to all work fine for me in my project) so would appreciate any help or comments.

### DIFF
--- a/Sources/Subjects/ReplaySubject.swift
+++ b/Sources/Subjects/ReplaySubject.swift
@@ -137,6 +137,7 @@ extension ReplaySubject {
 
         func forwardCompletionToBuffer(_ completion: Subscribers.Completion<Failure>) {
             demandBuffer?.complete(completion: completion)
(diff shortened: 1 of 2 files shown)


freak4pc:
Hey @jasdev - mind looking into this, as you know ReplaySubject well enough here? 

Thanks @jabeattie :)

jasdev (approved):
Thanks for PRing this @jabeattie! Looks good to me after a read through — really appreciate you adding a test to verify the fix. ✅

PR #92, opened 2021-06-09
🆕 Add a 'collect(until:)' operator

ferologics:
Useful for collecting elements when there the element count is unknown or does not fit any time grouping strategy.

### DIFF
--- a/Sources/Operators/CollectUntilTrigger.swift
+++ b/Sources/Operators/CollectUntilTrigger.swift
@@ -11,41 +11,128 @@ import Combine
 
 @available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
 public extension Publisher {
-    func collect<CollectionTrigger>(
-        until trigger: CollectionTrigger
-    ) -> AnyPublisher<[Output], Failure> where
-        CollectionTrigger: Publisher,
-        CollectionTrigger.Output == Void,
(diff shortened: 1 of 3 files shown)


freak4pc:
Hey @ferologics,
Thanks for the suggestion! I think it's a great idea but there are a few issues with the implementation.
Specifically, it is preferable that an operator doesn't use `sink` internally. Both because it's not a best practice, and because it automatically forces unlimited back pressure demand. 

It seems to me it would be more appropriate to either:
1. Find a way to compose other operators to make this work; or
2. Build a custom publisher (similar to others in this repo) to be the consumer for the events and the trigger

freak4pc:
Hey @ferologics, do you still want to pursue this operator?

ferologics:
Hey @freak4pc, yes! Thank you for the thoughtful response. AFAIK this will require a custom publisher implementation. I will take a look at the implementation of other publishers in the repo and update the code accordingly 🙂

ferologics:
@freak4pc I updated the PR with alternative implementation that avoids the `sink` on upstream. I haven't found a way to avoid the `sink` on trigger. Let me know if this is correct 🙌

VannDK (approved):

freak4pc:

ferologics:
all good @freak4pc :) cheers

PR #93, opened 2021-06-17
operators: add fromAsync and toAsync to bridge Combine and Structured…

twittemb:
This PR adds 3 operators on `Publishers` to allow the execution of async function or async sequence in the context a Combine Publisher

- fromAsync: 
   - wraps a non throwing async function into a Publisher
  
- fromThrowingAsync
   - wraps a throwing async function into a Publisher
   
- fromAsyncSequence
   - wraps an async sequence into a Publisher

Any improvement is welcome.

### DIFF
--- a/.github/workflows/tests.yml~
+++ b/.github/workflows/tests.yml~
@@ -28,8 +28,10 @@ jobs:
       - uses: actions/checkout@v2
       - name: Select Xcode 13
         run: sudo xcode-select -s /Applications/Xcode_13.0.app
+#      - name: Generate project
+#        run: make project
       - name: Run tests
-        run: set -o pipefail && xcodebuild -scheme CombineExt-Package -enableCodeCoverage YES -sdk ${{ matrix.sdk }} -destination "${{ matrix.destination }}" test | xcpretty -c -r html --output logs/${{ matrix.platform }}.html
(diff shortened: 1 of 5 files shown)


Sources/Operators/ToAsync.swift
```diff
@@ -0,0 +1,54 @@
...
+
+#if canImport(Combine)
+import Combine
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+public extension Future {
+    func toAsync() async -> Output where Failure == Never {
+        var subscriptions = [AnyCancellable]()
+
+        return await withUnsafeContinuation { [weak self] (continuation: UnsafeContinuation<Output, Never>) in
```
freak4pc, line 19: Why unsafe continuation? Seems like a classic checked throwing case here

twittemb: As discussed together, although withUnsafeContinuation seems "safe" to use with a Future, we'll go on with a "withCheckedContinuation" as we don't want to provide an API that could lead to an inconsistant state. That way, if there are multiple calls to "resume", there will be a runtime error.

twittemb:
@freak4pc I think the implementation is not too bad. The CI does not pass due to swift / Xcode versions.
Do you know if Github action are ready to run Xcode13/Swift 5.5/iOS15 ?

twittemb:
@freak4pc I think I will rework this in order to handle the back pressure. I guess it is possible to create a new Publisher that will handle the async execution and use a Sink to dump the result(s).

PR #95, opened 2021-07-12
Package.swift: bump swift-tools-version to 5.3 and hide tests dependencies

r-mckay:
Hello!

This commit is meant to hide CombineSchedulers from the project using CombineExt.

Before it, a project would import:
Screen Shot 2021-07-12 at 15 05 32

After this commit, a project now imports:
Screen Shot 2021-07-12 at 15 05 08

In order to do so, we bumped swift-tools-version to 5.3 and altered the test target expression.

Please note that I am not a Package.swift nor a SPM expert, comments and feedbacks will be very appreciated. 

Thank you!

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.3
 
 import PackageDescription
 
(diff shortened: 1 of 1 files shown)


r-mckay:
Hi! 
Just putting some light on this PR as it will help us with our software composition analysis by removing these two dependencies.
Is it possible to have someone look at it? @freak4pc, your opinion on this would be very appreciated.
Thanks!

freak4pc (approved):
Thanks! I must say I'm not an expert in SPM and I had thought the dependencies of a test target are hidden by default but that doesn't seem the case. Lets try going with this and hope its won't cause any breakage.

r-mckay:
Thanks, let's do it this way. Please let me know if there is anything. Thank you for your feedback.

PR #96, opened 2021-08-03
AllSatisfy operator

vitallica:
- Added an `AllSatisfy` operator.
- Added tests.
- Updated `README`.

This is my first contribution to this library, so please feel free to offer any feedback.


freak4pc:
Hey @vitallica ... It seems this operator already exists? 
https://developer.apple.com/documentation/combine/publishers/allsatisfy

vitallica:
Hey @freak4pc, thanks for looking at this PR.

The native operator does not handle situations where `Output` is of type `Collection`, since the `predicate` takes a value of type `Self.Output`. This operator has a `predicate` that takes a value of type `Output.Element`.

vitallica:
@freak4pc while investigating further, I found a case in which the compiler chooses the native operator rather than this one.
```swift
let name = ["John", "Jane", "Jim", "Jill", "Joan"]

Just(name)
  .allSatisfy { $0.count <= 4 }
  .sink { print("all short? \($0)") }
```
This returns `false`, because the array itself has a count of 5. However it's possible to force the overload like so:
```swift
Just(name)
  .allSatisfy { (name: String) in name.count <= 4 }
  .sink { print("all short? \($0)") }
```
This returns `true`, because all of the elements within the array have `count <= 4`.

In order to avoid this conflict, I think I should rename this operator to `AllSatisfyMany` to match the naming convention of `CombineLatestMany`, `MapMany`, and `MergeMany`. What do you think?

freak4pc:

PR #100, opened 2021-09-04
operators: add compactScan() and tryCompactScan()

twittemb:
This PR adds a compactScan() and a tryCompactScan() operators.

`compactScan()` transforms elements from the upstream publisher by providing the current element to a closure along with the last value returned by the closure. If the closure returns a nil value, then the accumulator won't change until the next non-nil upstream publisher value.

The `tryCompactScan()` version behaves the same but with a throwing closure.


freak4pc:

PR #106, opened 2021-09-26
operators: add fromAsync, fromThrowingAsync, fromAsyncSequence

twittemb:
This PR bridges structured concurrency with Combine by adding:
- Publishers.fromAsync(): wraps an async function call into an AnyPublisher<Output, Never>
- Publishers.fromThrowingAsync(): wraps a throwing async function call into an AnyPublisher<Output, Error>
- Publishers.fromAsyncSequence(): wraps an async sequence into an AnyPublisher<Output, Error>

The CI has been adapted to allow the usage of Xcode 13/Swift 5.5.

The SPM section has been removed since `xcodebuild` can now run swift packages (no Xcode project generate needed).

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/CombineExt.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/CombineExt.xcscheme
@@ -1,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "1300"
+   version = "1.3">
+   <BuildAction
+      parallelizeBuildables = "YES"
+      buildImplicitDependencies = "YES">
+      <BuildActionEntries>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "CombineExt"
+               BuildableName = "CombineExt"
+               BlueprintName = "CombineExt"
+               ReferencedContainer = "container:">
+            </BuildableReference>
+         </BuildActionEntry>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "NO"
+            buildForArchiving = "NO"
+            buildForAnalyzing = "YES">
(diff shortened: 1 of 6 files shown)


twittemb:
@freak4pc the code coverage does not seem happy, but I don't know why. The new code has a 100% coverage but the overall percentage seems to be lower than before.Do you know why ? (it's like that unit tests code coverage is taken into account)

markst:
is there an update on this?

Sources/Operators/FromAsync.swift
```diff
@@ -0,0 +1,191 @@
...
+//  CombineExt
+//
+//  Created by Thibault Wittemberg on 2021-06-15.
+//  Copyright © 2021 Combine Community. All rights reserved.
+//
+
+#if canImport(Combine)
+import Combine
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
```
markst, line 12: I believe this can be: 
```swift
#if compiler(>=5.5) && canImport(_Concurrency)
```
rather than forcing iOS15

https://www.swiftbysundell.com/special/swift-concurrency-backward-compatibility/

markst:
Seems `.subscribe(on: RunLoop.main)` has no effect:

```swift
Publishers.fromThrowingAsync(priority: TaskPriority.background, { () -> [EventEntity] in
    let result = try await self.useCase.invoke(forceRefresh: true)
    return result
})
.mapError { QatarErrorMap.other(message: $0.localizedDescription) }
.subscribe(on: RunLoop.main)
.eraseToAnyPublisher()
```

The `asyncThrowingFunction` is called on whatever thread the publisher was created from.

.swiftpm/xcode/xcshareddata/xcschemes/CombineExt.xcscheme
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
RomanPodymov, line 1: I think you should not commit this file.

RomanPodymov (changes requested):

.swiftpm/xcode/xcshareddata/xcschemes/CombineExt.xcscheme
markst: add this to gitignore:
```
# Ignore all schemes located in swiftpm
**/.swiftpm/**/xcschemes/
```

freak4pc:

PR #111, opened 2021-10-09
Add `Using` operator and tests.

danielt1263:
Adding the Using "operator" which I use quite a bit when working with RxSwift.


danielt1263:
Conflicts fixed. The operator is ready to be merged in with the next dot release.

PR #112, opened 2021-11-16
Properly cancel upstream work when using Materialize operator.  Closes #108

### LINKED ISSUE #108: Materialize prevents receiving cancel event
```
var cancellable  = publiser. handleEvents(receiveCancel: {
                Swift.print("cancel before materialise")
            })
            .materialize()
            .handleEvents(receiveCancel: {
                Swift.print("cancel before compact")
            }.sink{ _ in  }

cancellable.cancel()
```

prints 

```
cancel before compact
```

expected 

```
cancel before compact
cancel before materialise
```

jamieamazon:
This issue is easily reproducible by running the unit test I have added. The materialize operator will block any upstream work from being cancelled. The root cause seems to be due to a reference cycle where upstream has a reference to the sink, and the sink has a reference to upstream.  The cycle resolves after either a) upstream completes, or b) sink kills the upstream.  This is why simply nil'ing out the sink doesn't cancel upstream, because the sink deinit is not called.

### DIFF
--- a/Sources/Operators/Materialize.swift
+++ b/Sources/Operators/Materialize.swift
@@ -92,6 +92,7 @@ private extension Publishers.Materialize {
         }
 
         func cancel() {
+            sink?.cancelUpstream()
             sink = nil
         }
     }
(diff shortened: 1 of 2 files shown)


hebdavepaul0:
This fixes a memory leak in our codebase, what needs to happen to get it approved and merged?

freak4pc (approved):
Great catch, thank you!

PR #113, opened 2022-01-14
Add asSingle

stareta1202:
Support asSingle() on Publishers like RxSwift

### DIFF
--- a/CombineExt.xcodeproj/project.pbxproj
+++ b/CombineExt.xcodeproj/project.pbxproj
@@ -22,6 +22,7 @@
 /* End PBXAggregateTarget section */
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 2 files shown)


freak4pc (commented):
Hey, thank you for the idea.
I think this is a tricky change to add - 

`Single` is much more than a publisher. 
It republished events of a specific format (Result), has dedicated operators, checks and rules that don't allow working with multiple values etc.

Just adding the Publisher itself isn't really enough in this case /: 
If you're serious about this I'm open to accepting a larger PR that covers all of these edge cases but it might become quite broad.

WDYT?

freak4pc:
Sorry, "as-is" I don't think this is a good fit. 
If you'll look in RxSwift, `Single` has an entire suite of operators that work only on Single. If it's just a regular publisher and I can do non-Single things on it, it kinda loses the value of that guarantee.

PR #119, opened 2022-03-22
Implementation of FlatMapFirst operator

troupmar:
`FlatMapFirst` operator is a special case of `flatMap` operator.
 
Like `FlatMapLatest`, it only allows one inner publisher at a time. Unlike `FlatMapLatest`, it will not cancel an ongoing inner publisher.
Instead it ignores events from the source until the inner publisher is done. It creates another inner publisher only when the previous one is done.

Sources:
- https://danielt1263.medium.com/rxswifts-many-faces-of-flatmap-5764e6c2018c
- https://www.oreilly.com/library/view/mastering-reactive-javascript/9781786463388/298ee42c-d800-4228-87dd-a42d2ae6e820.xhtml
- https://github.com/ReactiveX/RxSwift/blob/1a1fa37b0d08e0f99ffa41f98f340e8bc60c35c4/RxSwift/Observables/Merge.swift

### DIFF
--- a/Sources/Operators/FlatMapFirst.swift
+++ b/Sources/Operators/FlatMapFirst.swift
@@ -1,0 +1,53 @@
+//
+//  FlatMapFirst.swift
+//  CombineExt
+//
+//  Created by Martin Troup on 22/03/2022.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+#if canImport(Combine)
+import Combine
+import Foundation
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+public extension Publisher {
+    /// The operator is a special case of `flatMap` operator.
+    ///
+    /// Like `flatMapLatest`, it only allows one inner publisher at a time. Unlike `flatMapLatest`, it will not cancel an ongoing inner publisher.
+    /// Instead it ignores events from the source until the inner publisher is done. It creates another inner publisher only when the previous one is done.
+    ///
+    /// - Returns: A publisher emitting the values of a single inner publisher at a time (until the inner publisher finishes).
+    func flatMapFirst<P: Publisher>(
+        _ transform: @escaping (Output) -> P
+    ) -> Publishers.FlatMap<Publishers.HandleEvents<P>, Publishers.Filter<Self>>
+    where Self.Failure == P.Failure {
+        var isRunning = false
+        let lock = NSRecursiveLock()
+
+        func set(isRunning newValue: Bool) {
+            defer { lock.unlock() }
+            lock.lock()
+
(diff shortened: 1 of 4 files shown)


Sources/Operators/FlatMapFirst.swift
```diff
@@ -0,0 +1,53 @@
...
+        let lock = NSRecursiveLock()
+
+        func set(isRunning newValue: Bool) {
+            defer { lock.unlock() }
+            lock.lock()
+
+            isRunning = newValue
+        }
+
+        return self
```
freak4pc, line 35: Interesting idea for an implementation! Also seems like a good test suite to cover it.
Can you remove the explicit self here?

troupmar: Sure, it should be gone...

Tests/FlatMapFirstTests.swift
```diff
@@ -0,0 +1,128 @@
...
+            .store(in: &cancellables)
+
+        testScheduler.advance(by: 2)
+
+        XCTAssertEqual(innerPublisherSubscriptionCount, 1)
+        XCTAssertEqual(innerPublisherCompletionCount, 1)
+        XCTAssertTrue(isUpstreamCompleted)
+    }
+
+    func test_flatMapFirst_error_upstream_skipping_flatMap() {
```
freak4pc, line 62: Can we change the naming to not use underscores? 
I believe that's not really common practice in Swift. 

Could also be a bit denser. Since we are in flatMapFirstTests, there's no need for `test_flatMapFirst` prefix. You can just call it `testErrorSkipsFlatmap`, or however you'd like. (This applies to the other tests too)

freak4pc (commented):
Sorry for taking so long! I'd like to merge this, just a quick note on the test naming and we can go ahead. @troupmar 

If you don't have the time let me know and I'll do my best to take care of it myself.

Tests/FlatMapFirstTests.swift
troupmar: I got it renamed!

freak4pc: seems likje a minor conflict to fix and then we can merge :) thank you.

troupmar: My bad, sorry. I resolved it.

freak4pc (approved):

freak4pc:
Still conflicted @troupmar :)

troupmar:
There were new conflicts introduced, I guess. But I resolved them as well! Still waiting for "1 workflow awaiting approval" ;).

freak4pc:
Thank you!

troupmar:
Thanks for the assist! :)

Issue #115, opened 2022-02-24
Unexplainable behavior of share-replay operator


troupmar:
Hello,

I have used `share(reply:)` operator in the past and just now I have encountered a corner case scenario, where it seems not to be working for some reason. I cannot say it is a problem of the `share(replay)` operator per se since I do not completely understand what and where the problem is. 

Here is a simple failing test:
```
  func testShareOperator() {
    let expectation = self.expectation(description: "")

    let publisher = Just("random-test-string")

    let sharedPublisher = publisher.share(replay: 1)

    let downstream1 = sharedPublisher
    let downstream2 = sharedPublisher

    Publishers.Zip(downstream1, downstream2)
      .sink { _ in
        expectation.fulfill()
      }
      .store(in: &self.subscriptions)

    waitForExpectations(timeout: 1)
  }
```

For some reason the `Zip` operator "freezes" and does not deliver the expected value. 

If the publisher is given a tiny delay, the test passes. I guess a thread hop is needed for the value to get through. However I do not understand why.
```
    let publisher = Just("random-test-string")
      .delay(for: 0.00000000001, scheduler: DispatchQueue.main)
```

`Zip` is not the only operator that does not work for me. I am facing the same problem if using `CombineLatest` instead. However the `Concatenate` operator, for example, does not have the same problem. I am guessing there is a problem with how (concurrently?) the `downstream` publishers are being fired in the operator. That is really my only clue. 

Maybe I am missing something basic and if so, I am sorry for opening the issue. But I have been really struggling to understand what is going on and any help is appreciated!

Best,
Martin

freak4pc:
I'm going to close this since it's super-old and I'm trying to clean up this repo. If this is still an issue / or something you'd like to pursue, feel free to comment and we can re-discuss. Thanks!

PR #120, opened 2022-03-22
Fix of ShareReplay operator for sequential run

troupmar:
**PROBLEM** (Open issue: https://github.com/CombineCommunity/CombineExt/issues/115)

The `share(replay:)` operator freezes when the upstream publisher runs sequentially. 
Following operations are called in this order:
1) `receive(subscriber:)` - replays buffer to the subscriber. At this point the buffer is empty since it has not received any values from the upstream publisher yet.
2) The sequential upstream sends its values. If `Just("random-value")` is used as an upstream publisher, it emits single value and finished event. The subject receives those via `send(_:)` and `send(completion:)` and forwards them to the subscriber. At this time `request(demand:)` has not yet been called on the subscriber. Demand is `zero` and thus the values are not received by the subscriber. 
3) `forwardCompletionToBuffer(_:)` is called on subscriber after `send(completion:)` is received on the subject. It calls `cancel()` which cancels the subscriber. The subscriber does not receive any values and finishes prematurely. 

**FIX** 

The replaying of the values happen only after the `request(demand:)` is called on the subscriber and thus the subscriber is ready to receive values. 
The `cancel()` on the subscriber is called only after the `request(demand:)` is called on the subscriber and thus the subscriber actually has a chance to receive values before being cancelled.

### DIFF
--- a/Sources/Subjects/ReplaySubject.swift
+++ b/Sources/Subjects/ReplaySubject.swift
@@ -41,17 +41,17 @@ public final class ReplaySubject<Output, Failure: Error>: Subject {
         let subscriptions: [Subscription<AnySubscriber<Output, Failure>>]
 
         do {
-          lock.lock()
-          defer { lock.unlock() }
+            lock.lock()
+            defer { lock.unlock() }
 
-          guard isActive else { return }
+            guard isActive else { return }
 
-          buffer.append(value)
-          if buffer.count > bufferSize {
-            buffer.removeFirst()
-          }
+            buffer.append(value)
+            if buffer.count > bufferSize {
+                buffer.removeFirst()
+            }
 
-          subscriptions = self.subscriptions
+            subscriptions = self.subscriptions
         }
 
         subscriptions.forEach { $0.forwardValueToBuffer(value) }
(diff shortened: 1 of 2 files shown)


Sources/Subjects/ReplaySubject.swift
```diff
@@ -41,17 +41,17 @@ public final class ReplaySubject<Output, Failure: Error>: Subject {
         let subscriptions: [Subscription<AnySubscriber<Output, Failure>>]
 
         do {
-          lock.lock()
-          defer { lock.unlock() }
+            lock.lock()
```
freak4pc, line 44: can you undo all of the re-formatting you did here please?

freak4pc:
Is this related to this in your opinion ?
https://github.com/CombineCommunity/CombineExt/pull/109/files

Sources/Subjects/ReplaySubject.swift
troupmar: I can do that. But it seems you are using Indent 4 everywhere and if you look into this file in the main branch, you will see that just this `do` scope has Indent 2. Let me know if that is what you want and I can change it back ;)

troupmar:
> Is this related to this in your opinion ?
> https://github.com/CombineCommunity/CombineExt/pull/109/files

Yes, I think the previously merged change (https://github.com/CombineCommunity/CombineExt/pull/86) caused one of the problems I am trying to fix here. However it is just a small part of the overall problem that I consider a bug and present the fix for here. 

I do not know the complete logic behind the https://github.com/CombineCommunity/CombineExt/pull/109 but I have tried to apply the presented changes and revert my logic with `isActive` within the `Subscription`. It seems my test is still passing, as well as the other ones, so as far as I am concerned it is fine. If you decide to merge the https://github.com/CombineCommunity/CombineExt/pull/109, I can rebase my change appropriately.

fabianmuecke:
I stumbled upon this PR, when I found my unit tests not working because of the `share(replay:) ` operator. I switched to the PR's branch and tried it. Now my unit tests work, but I get an infinite recursion (and therefore crash) when running my app. I'll try to find out, what's going on. Just wanted to let you know, that maybe there's still a problem with this.

troupmar:
> I stumbled upon this PR, when I found my unit tests not working because of the `share(replay:) ` operator. I switched to the PR's branch and tried it. Now my unit tests work, but I get an infinite recursion (and therefore crash) when running my app. I'll try to find out, what's going on. Just wanted to let you know, that maybe there's still a problem with this.

Hi, thanks for letting me know! We use the operator at some limited places and it seems to be working just fine. It would be cool if you managed to write a test that simulates the problem. Let me know if there is any progress on that ;)

fabianmuecke:
I haven't been able to come up with a test case, which reproduces our runtime issue, yet. It's somehow related to the `Multicast` triggering `request(demand:)` again synchronously, when receiving input, so the replay you added in that function gets called again, which calls `receive(input:)` on the `Multicast`, which in turn calls `request(demand:)` again and so on...

I noticed, the linked PR https://github.com/CombineCommunity/CombineExt/pull/109/files solves the same issue, this PR solves, without causing our runtime issues. The cause of the problem seems to be, that the `demandBuffer` gets cleared on completion. The other PR 109 fixes that, by not calling `cancel`. As `request(demand:)` does not call `replay` there, there's no infinite recursion. The test case you added, works on that other branch, too. From what I figured, PR 109 has not yet been merged, because it is missing a test case. So maybe, if your test case was added there, it could be merged, rendering this PR obsolete?

mRs-:
Hey there, thanks for the contribution for CombineExt!

From my understanding we should do a joint venture between #109 and #120. The Implementation from #109 looks more simple and from my understanding it would be better to not to clear the demandBuffer.

My Idea would be to use the implementation of #109 and the Tests from #120

freak4pc:
I agree with @mRs- that @jabeattie's solution is simpler. If your test suite passes for that PR it might be better to accept a more succinct change here unless there's a reason not to.

troupmar:
@fabianmuecke @mRs- @freak4pc Yes, I agree. I asked @jabeattie to [add the unit test in his PR](https://github.com/CombineCommunity/CombineExt/pull/109#issuecomment-1094766950).

mRs-:
I've created #124

freak4pc:
I'm closing this in favor of the merged #124.
Thanks all!

Issue #121, opened 2022-04-08
Memory leak in withLatestFrom operator


bharath2020:
Hello folks, 

We have found a memory leak issue in `withLatestFrom` operator which is very similar to the memory leak issue found [here](https://github.com/CombineCommunity/CombineExt/issues/85) in ShareReplay.

For better understanding, here is the code snippet

```
weak var weakSubject1: PassthroughSubject<Int, Never>?
weak var weakSubject2: PassthroughSubject<[Int], Never>?
autoreleasepool {
    let subject1 = PassthroughSubject<Int, Never>()
    let subject2 = PassthroughSubject<[Int], Never>()
    
    weakSubject1 = subject1
    weakSubject2 = subject2
    
    let subscritpion = subject1
        .withLatestFrom(subject2)
        .map { (a) -> AnyPublisher<Void, Never> in
            CurrentValueSubject<Void, Never>(()).eraseToAnyPublisher()
        }
        .sink { _ in

        }
    subject2.send([])
    subject1.send(1)
    subject2.send([])
}

weakSubject2 == nil // destroyed as expected.
weakSubject1 == nil // should have been destroyed but not
``` 

Expected: 
subject1 and subject2 must be destroyed after the autorelease pool got  drained

Actual:
subject1 is held in memory and becomes a zombie even after autorelease pool it belonged to was drained.

Solution:
It is very similar to how the share replay memory leak is taken care in this [PR](https://github.com/CombineCommunity/CombineExt/commit/547d11ac6a1f1090149e9724f303c16e7765f406#diff-d8695fcb7b3778a3467542060de9527a90fe258c4367d2e6464f98a6fad05facR95)

I am happy to put up a PR to fix the issue. Please let me know if there is a better way to handle this.

bharath2020:
@freak4pc Thanks for approving and merging the PR. When can I expect the next release? We would want to update our App to include this fix ASAP.

bharath2020:
@freak4pc Just bumping this one up. Can you please cut the release or is there a automated process to do so? We would want to release this ASAP as we are using this operator heavily.

freak4pc:
@bharath2020 You can simply lock your app to the specific commit. I'll try to find time to release soon, but you need to remember contributors don't work full time at cutting releases :) Since it's not an actual blocker there's nothing urgent in cutting the release.

bharath2020:
[...]

PR #122, opened 2022-04-11
fix memory leak of receiver while using withLatestFrom operator

bharath2020:
Fixing a memory leak in `withLatestFrom` operator as detailed in this issue https://github.com/CombineCommunity/CombineExt/issues/121


freak4pc (approved):
Nice catch, thank you so much!

PR #128, opened 2022-05-22
⚠️ Fix `map(to:)` name collision

danshevluk:
Hello everyone!

After the latest update `1.6.0` our project that uses `CombineExt` stopped compiling with a wired error message. After some digging, I figured out that in some cases Swift confuses `map(to:)` extension with `map` operator and uses closure as a constant itself.

It can be reproduced with a test case that I added:
```swift
let fooSubject = PassthroughSubject<Int, Never>()
let barSubject = PassthroughSubject<Int, Never>()

let combinedPublisher = Publishers.CombineLatest(fooSubject, barSubject)
   .map { fooItem, barItem in
       fooItem * barItem // Error here: "ambiguous use of operator '*'"
   }
// adding .eraseToAnyPublisher() will produce AnyPublisher<(Int, Int) -> Int, Never>

_ = combinedPublisher
    .map {
        "\($0)"
     }
     .sink {  
         print($0)
     }

fooSubject.send(5)
barSubject.send(6)
```

One possible solution is to rename this operator to `mapToValue`. This both makes sense semantically, fixes name collision, and is very similar to `mapToResult`.
Also, added more tests for updated `mapToValue` ✨

> ⚠️ Note: this is a breaking change, but it fixes a critical issue.

This is my first PR to CombineExt, please let me know what you think! ✌️

### DIFF
--- a/Sources/Operators/MapTo.swift
+++ b/Sources/Operators/MapTo.swift
@@ -14,8 +14,8 @@ public extension Publisher {
     /// Replace each upstream value with a constant.
     ///
     /// - Parameter value: The constant with which to replace each upstream value.
-    /// - Returns: A new publisher wrapping the upstream, but with output type `Result`.
-    func map<Result>(to value: Result) -> Publishers.Map<Self, Result> {
+    /// - Returns: A new publisher wrapping the upstream, but with output type `Value`.
+    func mapToValue<Value>(_ value: Value) -> Publishers.Map<Self, Value> {
         map { _ in value }
     }
 }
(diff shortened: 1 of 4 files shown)


Sources/Operators/MapTo.swift
```diff
@@ -14,8 +14,8 @@ public extension Publisher {
     /// Replace each upstream value with a constant.
     ///
     /// - Parameter value: The constant with which to replace each upstream value.
-    /// - Returns: A new publisher wrapping the upstream, but with output type `Result`.
-    func map<Result>(to value: Result) -> Publishers.Map<Self, Result> {
+    /// - Returns: A new publisher wrapping the upstream, but with output type `Value`.
+    func mapToValue<Value>(_ value: Value) -> Publishers.Map<Self, Value> {
```
freak4pc, line 18: I think it's worth renaming the file. Aside for that this looks good.

freak4pc (approved):

PR #129, opened 2022-05-22
Add `mapToVoid` operator

danshevluk:
After [fixing](https://github.com/CombineCommunity/CombineExt/pull/128) a `mapToValue` operator I got an idea to add `mapToVoid` operator. It's the most common use case for mapping publisher to a constant value. For example if you subscribe to `NotificationCenter` publisher and do not care about the value it produces:

```swift
NotificationCenter.default.publisher(for: UIApplication.userDidTakeScreenshotNotification)
    .mapToVoid()
   .sink {
       // Do something
   }
```

### DIFF
--- a/Sources/Operators/MapToValue.swift
+++ b/Sources/Operators/MapToValue.swift
@@ -18,5 +18,12 @@ public extension Publisher {
     func mapToValue<Value>(_ value: Value) -> Publishers.Map<Self, Value> {
         map { _ in value }
     }
+
+    /// Replace each upstream value with Void.
+    ///
+    /// - Returns: A new publisher wrapping the upstream and replacing each element with Void.
+    func mapToVoid() -> Publishers.Map<Self, Void> {
+        mapToValue(())
+    }
 }
 #endif
--- a/Sources/Operators/MapToValue.swift
+++ b/Sources/Operators/MapToValue.swift
@@ -23,7 +23,7 @@ public extension Publisher {
     ///
     /// - Returns: A new publisher wrapping the upstream and replacing each element with Void.
     func mapToVoid() -> Publishers.Map<Self, Void> {
-        mapToValue(())
+        map { _ in () }
     }
 }
 #endif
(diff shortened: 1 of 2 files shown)


freak4pc:
Hey, thanks for the idea ! 
I'm not entirely sure this operator is worthwhile. 

`map { _ in }`
`mapToVoid()`
`mapToValue(())`

We're not really saving so much letter count with this very specific method. 
If there's more demand for it we can happily add it, but otherwise It seems like something that might be better off as a specific extension to some codebase. 

WDYT?

danshevluk:
@freak4pc 

I had something similar in most projects that I worked on, so it feels like a common thing to use. I agree that it's not saving much letter space, but the main purpose of this extension is to explicitly mark muting the upstream. 

Practical example. Imagine that you want to refresh some app data in one of 4 cases: 
- data model updated
- refresh control triggered
- if some timer passed
- if the app became active

Here is how I would implement this using `mapToVoid()`. Looks cleaner than with `.map { _ in }` or `mapToValue(())`
```swift
Publishers.Merge4(
   $addressInfo.removeDuplicates().mapToVoid(),
   refreshControlPublisher.mapToVoid(),
   refreshTimerPublisher.mapToVoid(),
   NotificationCenter.default.publisher(for: UIApplication.didBecomeActiveNotification).mapToVoid()
)
// Published value type here is Void
.throttle(for: 4, scheduler: DispatchQueue.main, latest: true)
.sink {
   // schedule app data refresh
}
.store(in: &cancellables)
```

Is's more of a convenience method, but I feel like it's a good fit for Combine`Ext`. Your call! 😃

Sources/Operators/MapToValue.swift
```diff
@@ -18,5 +18,12 @@ public extension Publisher {
     func mapToValue<Value>(_ value: Value) -> Publishers.Map<Self, Value> {
         map { _ in value }
     }
+
+    /// Replace each upstream value with Void.
+    ///
+    /// - Returns: A new publisher wrapping the upstream and replacing each element with Void.
+    func mapToVoid() -> Publishers.Map<Self, Void> {
+        mapToValue(())
```
freak4pc, line 26:
```suggestion
        map { _ in () }
```

let's make it independent

freak4pc (approved):
I'll go for it :) one fix

danshevluk:
@freak4pc do you have any further suggestions? To me looks like it's good to merge, can you do that? 🙏

PR #130, opened 2022-05-25
combineLatest & zip: make scalable using divide-and-conquer.

fischman-bcny:
Previously combineLatest's implementation would trigger a stack overflow on 15000
publishers because the implementation combined its inputs linearly.
This commit replaces that with repeated (hierarchical) 4-way merges,
which lets the same scaled test scenario go to 1e7 publishers with no
issues (other than taking 28s to run the test on my M1 Pro, hence the
choice to commit the test with only 1e5 publishers). Test execution
wall-time scales linearly with publisher count.

Re: "4-way merge":
- Why not use a higher value? 4 is the highest arity offered by
  Combine's Publisher.combineLatest.
- Why not use a lower value? Using 2-way merges instead of 4-way still
  lets the 1e7 test case pass, but is ~60% slower (1e7 publishers take
  45s insted of 28s).

All of the above applies remarkably similarly to zip, also changed in the same way in this PR.

### DIFF
--- a/Sources/Operators/CombineLatestMany.swift
+++ b/Sources/Operators/CombineLatestMany.swift
@@ -46,13 +46,10 @@ public extension Collection where Element: Publisher {
     /// - returns: A type-erased publisher with value events from each of the inner publishers `combineLatest`’d
     /// together in an array.
     func combineLatest() -> AnyPublisher<[Element.Output], Element.Failure> {
-        if isEmpty { return Empty().eraseToAnyPublisher() }
-
         var wrapped = map { $0.map { [$0] }.eraseToAnyPublisher() }
         while wrapped.count > 1 {
             wrapped = makeCombinedQuads(input: wrapped)
         }
-
         return wrapped.first?.eraseToAnyPublisher() ?? Empty().eraseToAnyPublisher()
     }
 }
--- a/Sources/Operators/CombineLatestMany.swift
+++ b/Sources/Operators/CombineLatestMany.swift
@@ -48,31 +48,53 @@ public extension Collection where Element: Publisher {
     func combineLatest() -> AnyPublisher<[Element.Output], Element.Failure> {
         if isEmpty { return Empty().eraseToAnyPublisher() }
 
-        // Given an array of publishers, return an array of the result of combineLatest'ing each successive foursome of
-        // input publishers (or fewer, for the remainder of count%4).
-        func makeCombinedQuads(input: [AnyPublisher<[Element.Output], Element.Failure>]) -> [AnyPublisher<[Element.Output], Element.Failure>] {
-            sequence(state: input.makeIterator(), next: { it in it.next().map { ($0, it.next(), it.next(), it.next()) } })
-            .map { quad in
-                guard let second = quad.1 else { return quad.0 }
-                guard let third = quad.2 else { return quad.0.combineLatest(second)
-                        .map { $0.0 + $0.1 }
-                        .eraseToAnyPublisher()
-                }
-                guard let fourth = quad.3 else { return quad.0.combineLatest(second, third)
-                        .map { $0.0 + $0.1 + $0.2 }
-                        .eraseToAnyPublisher()
-                }
-                return quad.0.combineLatest(second, third, fourth)
-                    .map { $0.0 + $0.1 + $0.2 + $0.3 }
-                    .eraseToAnyPublisher()
-            }
-        }
-
         var wrapped = map { $0.map { [$0] }.eraseToAnyPublisher() }
         while wrapped.count > 1 {
             wrapped = makeCombinedQuads(input: wrapped)
         }
-        return wrapped[0].eraseToAnyPublisher()
+
+        return wrapped.first?.eraseToAnyPublisher() ?? Empty().eraseToAnyPublisher()
+    }
+}
+
+// MARK: - Private helpers
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+/// CombineLatest an array of input publishers in four-somes.
+///
+/// - parameter input: An array of publishers
+private func makeCombinedQuads<Output, Failure: Swift.Error>(
+    input: [AnyPublisher<[Output], Failure>]
+) -> [AnyPublisher<[Output], Failure>] {
+    // Iterate over the array of input publishers in steps of four
+    sequence(
+        state: input.makeIterator(),
+        next: { it in it.next().map { ($0, it.next(), it.next(), it.next()) } }
+    )
+    .map { quad in
+        // Only one publisher
+        guard let second = quad.1 else { return quad.0 }
+
+        // Two publishers
+        guard let third = quad.2 else {
+            return quad.0
+                .combineLatest(second)
+                .map { $0.0 + $0.1 }
+                .eraseToAnyPublisher()
+        }
+
+        // Three publishers
+        guard let fourth = quad.3 else {
+            return quad.0
+                .combineLatest(second, third)
+                .map { $0.0 + $0.1 + $0.2 }
+                .eraseToAnyPublisher()
+        }
+
+        // Four publishers
+        return quad.0
+            .combineLatest(second, third, fourth)
+            .map { $0.0 + $0.1 + $0.2 + $0.3 }
+            .eraseToAnyPublisher()
     }
 }
 #endif
--- a/Sources/Operators/CombineLatestMany.swift
+++ b/Sources/Operators/CombineLatestMany.swift
@@ -18,17 +18,9 @@ public extension Publisher {
     ///
     /// - returns: A type-erased publisher with value events from `self` and each of the inner publishers `combineLatest`’d
     /// together in an array.
-    func combineLatest<Others: Collection>(with others: Others)
-        -> AnyPublisher<[Output], Failure>
-        where Others.Element: Publisher, Others.Element.Output == Output, Others.Element.Failure == Failure {
-        let seed = map { [$0] }.eraseToAnyPublisher()
-
-        return others.reduce(seed) { combined, next in
-            combined
-                .combineLatest(next)
-                .map { $0 + [$1] }
-                .eraseToAnyPublisher()
-        }
+    func combineLatest<Others: Collection>(with others: Others) -> AnyPublisher<[Output], Failure>
+        where Others.Element == Self, Others.Element.Output == Output, Others.Element.Failure == Failure {
+            ([self] + others).combineLatest()
     }
 
     /// Projects `self` and a `Collection` of `Publisher`s onto a type-erased publisher that chains `combineLatest` calls on
@@ -38,9 +30,7 @@ public extension Publisher {
     ///
     /// - returns: A type-erased publisher with value events from `self` and each of the inner publishers `combineLatest`’d
     /// together in an array.
-    func combineLatest<Other: Publisher>(with others: Other...)
-        -> AnyPublisher<[Output], Failure>
-        where Other.Output == Output, Other.Failure == Failure {
+    func combineLatest(with others: Self...) -> AnyPublisher<[Output], Failure> {
         combineLatest(with: others)
     }
 }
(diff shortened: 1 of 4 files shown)


freak4pc:
Thanks for the PR! 
What's the use case you have trying to CombineLatest so many publishers ? 
It seems interested but might be heavily over-optimized with losing the current code clarity, so I want to make sure there's a clear value to this.

Also, with the existing implementation, how many "long seconds" does it take?

fischman-bcny:
> Thanks for the PR! 

Thanks for the quick look!

> What's the use case you have trying to CombineLatest 100k publishers ?

1e5 is overkill for my use-case, but the test in question is super-light on what each publisher does.
Context on this PR is that I had attempted to replace uses of a home-grown (much longer but incomplete) version of combineLatest with CombineExt's version in a large (but closed-source, sorry) project and got push-back during review that CombineExt's version had [known](https://jasdev.me/notes/variadic-zip-cash) scaling limits, and a worry that "real" (much heavier-weight than Just(2) :) ) publishers in the project in question might trigger this limit with "only" hundreds of publishers. This PR is an attempt to address that push-back :)

BTW this exchange also prompted me to binary-search a tighter bound; turns out the current impl crashes at 14474 publishers. Updated PR description to reflect this.

> It seems interested but might be heavily over-optimized with losing the current code clarity, so I want to make sure there's a clear value to this.

ISTM that any API that accepts a Collection and turns it into unoptimiz{ed,able} tail-called recursion is fraught, because an array with 10k elements isn't something that screams "huge! consider the implications!" to me. (unlike, say, an array with 1e9 elements).
Turning your question around: would it be better to keep the current implementation of combineLatest and add a comment saying "consider the scaling implications on collections of more than N publishers" for some explicit value of N?

> Also, with the existing implementation, how many "long seconds" does it take?

Thanks for asking and making me look! It took something like 20-40s to go from starting the test to showing the crash in Xcode, but because the test would pause at the crash (in lldb) I didn't have an exact duration. Attempting to find that duration by running the test outside lldb revealed that it's actually super-fast, too, and that it's lldb's generation of a backtrace and Xcode displaying it that was taking all those "long seconds". Elided that claim from the PR's description.

Sources/Operators/CombineLatestMany.swift
```diff
@@ -18,17 +18,9 @@ public extension Publisher {
...
-
-        return others.reduce(seed) { combined, next in
-            combined
-                .combineLatest(next)
-                .map { $0 + [$1] }
-                .eraseToAnyPublisher()
-        }
+    func combineLatest<Others: Collection>(with others: Others) -> AnyPublisher<[Output], Failure>
+        where Others.Element == Self, Others.Element.Output == Output, Others.Element.Failure == Failure {
+            ([self] + others).combineLatest()
```
fischman-bcny, line 23: @freak4pc @jasdev How important are this func and the next? This PR changes the typing slightly to be more restrictive (self and the combined-with publishers must share a type, instead of simply both being Publishers that share Output & Failure types), but apparently there isn't dedicated test coverage for them, and double-so for not having test coverage of different types of publishers being combined :)
The code-paths that would get here from the old impl would already enforce this same-type requirement (b/c "self" here is in the Collection there), so this seems fine to me.

freak4pc: We'd want to keep the existing surface area (meaning existing funds and return types), as otherwise it would be breaking

freak4pc:
> > Thanks for the PR!
> 
> Thanks for the quick look!
> 
> > What's the use case you have trying to CombineLatest 100k publishers ?
> 
> 1e5 is overkill for my use-case, but the test in question is super-light on what each publisher does. Context on this PR is that I had attempted to replace uses of a home-grown (much longer but incomplete) version of combineLatest with CombineExt's version in a large (but closed-source, sorry) project and got push-back during review that CombineExt's version had [known](https://jasdev.me/notes/variadic-zip-cash) scaling limits, and a worry that "real" (much heavier-weight than Just(2) :) ) publishers in the project in question might trigger this limit with "only" hundreds of publishers. This PR is an attempt to address that push-back :)
> 
> BTW this exchange also prompted me to binary-search a tighter bound; turns out the current impl crashes at 14474 publishers. Updated PR description to reflect this.
> 
> > It seems interested but might be heavily over-optimized with losing the current code clarity, so I want to make sure there's a clear value to this.
> 
> ISTM that any API that accepts a Collection and turns it into unoptimiz{ed,able} tail-called recursion is fraught, because an array with 10k elements isn't something that screams "huge! consider the implications!" to me. (unlike, say, an array with 1e9 elements). Turning your question around: would it be better to keep the current implementation of combineLatest and add a comment saying "consider the scaling implications on collections of more than N publishers" for some explicit value of N?
> 
> > Also, with the existing implementation, how many "long seconds" does it take?
> 
> Thanks for asking and making me look! It took something like 20-40s to go from starting the test to showing the crash in Xcode, but because the test would pause at the crash (in lldb) I didn't have an exact duration. Attempting to find that duration by running the test outside lldb revealed that it's actually super-fast, too, and that it's lldb's generation of a backtrace and Xcode displaying it that was taking all those "long seconds". Elided that claim from the PR's description.

Thanks for the detailed response.

Here's I'm feeling a bit indecisive, I want to share a bit of my thinking process :) 

I'll start from the end here:
>  because an array with 10k elements isn't something that screams "huge! consider the implications!" to me.

I'm having a hard time with this sentence. 10k elements in an array, sure. But 10k Publishers, where each publisher can be some work, and each change to any of the 10,000 elements would emit a recombination of all 10,000 elements — that isn't really a common use case. To be honest, I've been working with Reactive Tech for ... 8 years, and I've never had to CombineLatest 10,000 (or even 1000) different Observables/Publishers.

The code you wrote is more effective, for sure, but it loses a lot of the readability — This isn't about being "pretty". It's more about "can a different contributor come in easily and feel comfortable working on this", and I'm worried it's probably a "No", which is a relatively big sacrifice.

So my current thinking is that I'm not entirely sure this change is worth the cost. I'd love to get @jasdev's opinion because from one side he understands that balance IMO, and from the other end he hit these performance limits himself, too. 

I will say the use-case with "Zip" is actually much more obvious - zipping 100,000 requests is extreme, but I can think of some use cases. And once they're done, it's done. CombineLatest is basically never-ending with that amount of publishers, so I'm very curious about the use case here.

Anyways, I'm leaving this open, and I hope @jasdev will be able to chime in. I'll try to ping him directly if not.

Again, thank you so much for the optimization, it's a really interesting discussion which is very hard to find a good balance on 🙏

fischman-bcny:
Thanks for the thoughts, @freak4pc.
Funny you mention zip b/c I just got done applying the same treatment to it (I had forgotten @jasdev's blog post implicated it as well until I cited it to you above :) ). Added to this PR. 
Looking forward to hearing Jasdev's thoughts.

jasdev:
Sorry for the delayed reply here and love all the discussion!

My lean is to land this change. With concurrency being added directly into Swift, the writing is on the wall that Combine won’t be receiving updates anymore. Which, in my opinion, means it probably isn’t worth it to try and write custom variadic `zip` and `combineLatest` publishers à la the precedent [in RxSwift](https://github.com/ReactiveX/RxSwift/blob/a580d07ed002217fd91d8446c3a852486e9beefa/RxSwift/Observables/Zip%2BCollection.swift#L37-L169), [OpenCombine](https://github.com/OpenCombine/OpenCombine/pull/119/files#diff-8a3f4fa06e11ee74845c843c37cb874d08dc98b3a79f4c08fbe08ad85fc88f93), et al.

So, if writing a custom publisher — instead of the existing composed approach — is too much effort for a library (Combine) that likely won’t be updated and the existing approach has perf. bottlenecks, then @fischman-bcny’s PR seems like a nice middle ground to raise the perf. ceiling.

> …it’s more about “can a different contributor come in easily and feel comfortable working on this,” and I’m worried it’s probably a “No,” which is a relatively big sacrifice.

After this change, the only updates I can realistically see for these operators is moving them over to a custom publisher, which for the above reason, I don’t think is worth the effort, unless someone wanted to work on it as an learning exercise. 😄

—

tl;dr: I’m pro-landing this change. ✅

Sources/Operators/CombineLatestMany.swift
fischman-bcny: Fair enough. Figured out how to preserve the previous signatures.

Sources/Operators/CombineLatestMany.swift
```diff
@@ -53,19 +43,33 @@ public extension Collection where Element: Publisher {
...
+
+        // Given an array of publishers, return an array of the result of combineLatest'ing each successive foursome of
+        // input publishers (or fewer, for the remainder of count%4).
+        func makeCombinedQuads(input: [AnyPublisher<[Element.Output], Element.Failure>]) -> [AnyPublisher<[Element.Output], Element.Failure>] {
+            sequence(state: input.makeIterator(), next: { it in it.next().map { ($0, it.next(), it.next(), it.next()) } })
+            .map { quad in
+                guard let second = quad.1 else { return quad.0 }
+                guard let third = quad.2 else { return quad.0.combineLatest(second)
+                        .map { $0.0 + $0.1 }
+                        .eraseToAnyPublisher()
```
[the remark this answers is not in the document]
fischman-bcny: (discussed elsewhere and realized @jasdev was thinking these erasures were being applied to second/third/fourth but actually they're being applied to the result of the .map and so are indeed necessary to conform to makeCombinedQuads's signature (returning an array of AnyPublisher)

fischman-bcny:
@freak4pc back to you for an executive decision given @jasdev's input above, I think.

freak4pc:
@fischman-bcny I'll go with the democratic decision and will accept the solution. Thanks for the patience!
We have a company trip tomorrow so I won't be available until Thursday but I'll review when I'm back.

fischman-bcny:
@freak4pc hope you had a good trip :) 
Ping on this PR whenever you get a minute.

freak4pc:
LGTM! I've made a few minor clean-ups to make it more readable, I'll let CI run and marge it.

freak4pc (approved):

Sources/Operators/CombineLatestMany.swift
```diff
@@ -48,31 +48,53 @@ public extension Collection where Element: Publisher {
...
-            }
-        }
-
         var wrapped = map { $0.map { [$0] }.eraseToAnyPublisher() }
         while wrapped.count > 1 {
             wrapped = makeCombinedQuads(input: wrapped)
         }
-        return wrapped[0].eraseToAnyPublisher()
+
+        return wrapped.first?.eraseToAnyPublisher() ?? Empty().eraseToAnyPublisher()
```
fischman-bcny, line 56: if you're going to do this then l.49 is no longer needed :)

Sources/Operators/ZipMany.swift
```diff
@@ -42,31 +42,51 @@ public extension Collection where Element: Publisher {
...
-                    .eraseToAnyPublisher()
-            }
-        }
-
         var wrapped = map { $0.map { [$0] }.eraseToAnyPublisher() }
         while wrapped.count > 1 {
             wrapped = makeZippedQuads(input: wrapped)
         }
-        return wrapped[0].eraseToAnyPublisher()
+        return wrapped.first?.eraseToAnyPublisher() ?? Empty().eraseToAnyPublisher()
```
fischman-bcny, line 49: ditto

fischman-bcny (commented):
@freak4pc FYI, will let you apply or ignore these comments as you see fit.

Sources/Operators/ZipMany.swift
freak4pc: Good point, mind removing the lines at 49 and I'll merge it?

fischman-bcny: Done.

The-Igor:
Hi, I've been playing with the topic, slightly different take on this https://github.com/The-Igor/d3-network-service/blob/main/Sources/d3-network-service/extension/combine/Zipper.swift
[fischman-bcny](https://github.com/fischman-bcny) Interesting what your test will say in terms of performance

The-Igor:
Hi I've been playing with the topic, slightly diff take on this https://github.com/The-Igor/d3-network-service/blob/main/Sources/d3-network-service/extension/combine/Zipper.swift

fischman-bcny:
> Hi I've been also experimenting with the topic, slightly diff take on this https://github.com/The-Igor/d3-network-service/blob/main/Sources/d3-network-service/extension/combine/Zipper.swift [fischman-bcny](https://github.com/fischman-bcny) It's interesting what the test will show in terms of performance?!

@The-Igor our implementations appear to do the same thing except that mine uses a while loop where yours uses recursion to repeatedly process quads of inputs.

The-Igor:
I'm not trying to compete I'm trying to cooperate😊 It's really interesting is the difference in terms of performance.  Sometimes slightly less significant lead to better performance. As I got all the line is how to improve the current solution as much as possible.

fischman-bcny:
@The-Igor No offense taken.

Inlining your impl into the test file, bumping numPublishers to 5e6+1, and cloning testZipAtScale to suffix with a `2` shows your impl is just a smidge slower in Debug:
```
 Test Case '-[CombineExtTests.ZipManyTests testZipAtScale]' passed (17.278 seconds).
 Test Case '-[CombineExtTests.ZipManyTests testZipAtScale2]' passed (17.682 seconds).
```
but nontrivially faster in Release:
```
Test Case '-[CombineExtTests.ZipManyTests testZipAtScale]' passed (16.262 seconds).
Test Case '-[CombineExtTests.ZipManyTests testZipAtScale2]' passed (13.185 seconds).
```
(note that if you want to run the tests in Release you'll have to comment out ReplaySubjectTests.swift's contents)

My guess is that this is due to your code using arrays for the quads (and the `Array.count` being faster than the piece-meal `guard`s I have for nils in my (tuple-based) quads.
I'll let @freak4pc & @jasdev speak to whether the increased performance is compelling for the CombineExt project.
(I'm just a visitor here, motivated to make these operators not fall over in the presence of large inputs, and less motivated by performance optimizations for these scenarios)

Happy to do more work in this area if it's compelling for the project. 
Will proceed with this PR as-is to unblock downstream work (outside this project).

fischman-bcny:
@freak4pc merge plz? (I don't seem to have permission to)

PR #133, opened 2022-07-12
Fixes for the xcodeproj file

jasdeepsaini:
Fixes to get the project and tests to build:
- Deleted the missing `MapTo` and `MapToTests` files.
- Change the deployment target for the CombineExtTests target to 13.0 so the latest version of CombineSchedulers can be used.
- Removed CombineSchedulers from the CombineExt target and added it to CombineExtTests. This allows the project to build since CombineExt supports iOS 10.0 and CombineSchedulers requires iOS 13.
- Updated the CombineSchedulers package to 0.5.3 to match the Package file.
- The Package.resolved file automatically got updated to v2 since I am using Xcode 13.

Project cleanup
- Deleted the missing `Carthage` and `CHANGELOG.MD` files.

Missing library and test files added to the project:
- `MapToValue.swift`
- `MapToValueTests.swift`
- `MapToResult.swift`
- `MapToResultTests.swift`
- `Enumerated.swift`
- `EnumeratedTests.swift`


.gitignore
```diff
@@ -3,4 +3,4 @@ Packages
 xcuserdata
 DerivedData/
 Carthage
-CombineExt.framework.zip
\ No newline at end of file
+CombineExt.framework.zip
```
jasdeepsaini, line 6: I think the `.DS_Store` file should be added to `.gitignore`, but I didn't make that change in this PR. Is there a reason to commit this file?

freak4pc:
I'll tryto find some time to review this - I'll start by saying I'd prefer keeping the deployment target low (iOS 10 or 11), because thjere could be some older codebases that don't necessarily use Combine all over the place but can still benefit from using CombineExt in _some_ places that are guarded by availability clauses.

jasdeepsaini:
> I'll tryto find some time to review this - I'll start by saying I'd prefer keeping the deployment target low (iOS 10 or 11), because thjere could be some older codebases that don't necessarily use Combine all over the place but can still benefit from using CombineExt in _some_ places that are guarded by availability clauses.

The deployment target is an issue if you want to use the latest version of Combine-Schedulers which requires iOS 13. With the lower deployment target, the library and unit tests won't build.  The Package.swift file has a minimum iOS version of 13 so older projects won't be able to pull in this package.

freak4pc:
Anyways if that's the case we can leave the old version of combine-schedulers, It's just a test utility so I'm not sure it matters too much.

jasdeepsaini:
> Anyways if that's the case we can leave the old version of combine-schedulers, It's just a test utility so I'm not sure it matters too much.

I reverted the deployment target of CombineExt to iOS 10.0 and changed the deployment target of CombineExtTests to iOS 13.0.  Then I moved the CombineSchedulers package from CombineExt to CombineExtTests.

This should do what you wanted while allowing tests to be run if CombineExt is modified in Xcode instead of the command line.

freak4pc:
LGTM. Let's just fix the conflicts :) 
(Or add DS_Store to .gitignore which would be even better)

.gitignore
freak4pc: Would be great to add it ! Thanks

jasdeepsaini:
> LGTM. Let's just fix the conflicts :)
> (Or add DS_Store to .gitignore which would be even better)

I added .DS_Store to .gitignore.

.gitignore
jasdeepsaini: Done

PR #140, opened 2022-08-04
Fix memory leak in WithLatestFrom

### LINKED ISSUE #139: Memory Leak in ReplaySubject, WithLatestFrom, and anything else that uses the DemandBuffer
**[Repro Video](https://user-images.githubusercontent.com/2772733/182712355-08b548ee-64f2-4b43-8cc7-ba3783bdc084.mov)**
**[Sample Project](https://github.com/CombineCommunity/CombineExt/files/9255107/BugTest.zip)**

This project is setup with `ViewController` -> `ChildViewController` -> `GrandChildViewController`

The leak occurs only when `ChildViewController` sinks to the `postsSubject`

It seems that if a publisher is bound somewhere else, and then that same publisher is used with a `withLatestFrom` a retain cycle is introduced and Combine internally crashes

Seems like the `DemandBuffer` doesn't correctly forward completion information upstream if the upstream Subscriber is being used somewhere else which is why we are crashing in WithLatestFrom and share(replay: 1)

TTOzzi:
Resolves #139 

The withLatestFrom stream has been completed, but it has been confirmed that `otherSubscription` is remaining and causing memory leaks.
Modified it to call `cancel` when `WithLatestFrom`'s `Subscription` is deinitialized.

### DIFF
--- a/Sources/Operators/WithLatestFrom.swift
+++ b/Sources/Operators/WithLatestFrom.swift
@@ -234,6 +234,8 @@ private extension Publishers.WithLatestFrom {
         sink = nil
         otherSubscription?.cancel()
     }
+      
+    deinit { cancel() }
   }
 }
 #endif
--- a/Tests/WithLatestFromTests.swift
+++ b/Tests/WithLatestFromTests.swift
@@ -125,6 +125,29 @@ class WithLatestFromTests: XCTestCase {
         XCTAssertNil(weakSubject2)
     }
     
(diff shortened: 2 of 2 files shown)


adlerj:
This fix works in the sample project I made

freak4pc:
Awesome find and fix, thanks for this and for the extra test, too! 🤩

freak4pc (approved):

PR #141, opened 2022-08-17
Add bind(to: ) operator

cefiro666:
I would like to suggest this operator which copies the similar operator from the RxSwift. If you think this is a good idea, I need help writing clear documentation for the method.

### DIFF
--- a/Sources/Operators/Bind.swift
+++ b/Sources/Operators/Bind.swift
@@ -1,0 +1,27 @@
+//
+//  Bind.swift
+//  CombineExt
+//
+//  Created by Vitaly Banik on 17/08/2022.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+#if canImport(Combine)
+import Foundation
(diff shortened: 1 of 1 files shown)


Sources/Operators/Bind.swift
```diff
@@ -0,0 +1,27 @@
...
+    
+    func bind<T: Subject>(to subject: T) -> AnyCancellable
+    where T.Output == Output,
+          T.Failure == Failure {
+              
+        self.sink { [weak subject] completion in
+            subject?.send(completion: completion)
+        } receiveValue: { [weak subject] value in
+            subject?.send(value)
+        }
```
freak4pc, line 24: I think you just want https://developer.apple.com/documentation/combine/publisher/subscribe(_:)-3fk20

cefiro666: Oh. Apparently I don't know very well Combine =( Thank you!

PR #142, opened 2022-08-17
Allow editing the value property of the CurrentValueRelay

cefiro666:
I would like to be able to edit the value property of the CurrentValueRelay, as is possible for CurrentValueSubject


freak4pc:

PR #144, opened 2022-09-06
Amb: forward cancel to the inner publishers

melle:
We noticed that cancelling an Amb publisher does not cancel any of the inner publishers.
It looks like nil-ing firstSink / secondSink should do the job, but it does not work. Adding
explicit cancel fixes the issue, but I’m not sure if this is the right solution.


luizmb:
Moving discussion from Twitter to here: should `Sink` call cancelUpstream (https://github.com/CombineCommunity/CombineExt/blob/main/Sources/Common/Sink.swift#L107) upon deinit, the same way AnyCancellable does? The risk of not doing, is making the same assumption that setting it to nil will cleanup the resources (while in fact keeps an orphan subscription). The risk of changing Sink is potentially breaking other operators.

luizmb:
Moving discussion from Twitter.
Why Sink.deinit is not calling cancelUpstream automatically (https://github.com/CombineCommunity/CombineExt/blob/main/Sources/Common/Sink.swift#L113)?

Maybe it has a reference cycle and deinit is actually not called?

melle:
I looked into it: Sink's deinit is not called. That explains why the subscriptions are not cancelled. However, I could not find any explanation for the reference being kept alive. I added the cancel to the decision code as well including another test. 

The explicit cancel fixes _our_ problem, but shadows the underlying reference cycle problem. Both tests should pass, when the cancel() calls are removed and the reference cycle has been fixed.

freak4pc:

Issue #145, opened 2022-09-16
Documentation Building Failing on Xcode 14.0


nebsta:
Our project recently updated to Xcode 14, however when we try building the documentation, we are getting the following errors:

![Screenshot 2022-09-16 at 15 06 14](https://user-images.githubusercontent.com/94231771/190645701-eb8b4011-a87f-4ead-92be-22ad61fc34d8.png)

We're targeting iOS 15, and I have tried with versions 1.0.0 and 1.8.0 of CombineExt. Running Monterey 12.6. Also just tried the usual stuff of clearing the package cache, Derived Data etc.

I saw you guys posted an issue to the swift repo a little while ago about this issue, but I'm wondering if a fix for this is still in progress or if there is something else I'm missing? https://github.com/apple/swift/issues/60771

freak4pc:
I think they fixed it [in this PR](https://github.com/apple/swift/pull/61049), but unfortunately there's nothing we can do proactively from my side (that I know of). 

The culprit is probably in combine-schedulers itself, which re-exports Foundation (which funnily enough we don't even use aside for tests). 

You could possibly fork this repo and remove the tests, which should let you build the docs for it as a "workaround".

Hope this help, closing for now :)

jshier:
@freak4pc `combine-schedulers` has been updated to not export Foundation, so if you update to the latest version that issue should be fixed.

freak4pc:
That's great! @jasdev any chance you'll have a moment to take care of it?
If not, let me know and I'll do it.

jasdev:
https://github.com/CombineCommunity/CombineExt/pull/146

nebsta:
Thanks heaps for the quick response @jasdev + @freak4pc! Do you know when the next release is schedule for that will contain this fix?

freak4pc:
@nebsta I'd like to add 1-2 more fixes to it so I hope by Friday if that makes sense :)

nebsta:
@freak4pc No problem! Thanks again for jumping on this fix so quickly :)

PR #147, opened 2022-09-28
Remove stray exported import of Foundation.

jasdev:
Follow up PR for https://github.com/CombineCommunity/CombineExt/issues/145


jasdev:
(cc @freak4pc 🙏🏽)

nebsta:
When can we expect to have this merged into main? We're still having issues with building documentation as a result of the Foundation import, so it would be nice to get this in

jasdev:
Defer to @freak4pc since I don't have merge permissions. 🙏🏽

PR #156, opened 2023-04-14
AnyPublisher extensions for Just, Fail and Empty

### LINKED ISSUE #155: .just, .empty and .fail on AnyPublisher
I was wondering if there was a reason to not include in this (awesome) library some utility methods to simplify usage of Just, Fail and Empty, which requires an `eraseToAnyPublisher()` in so many scenarios.

I was thinking about something like this
```swift
public extension AnyPublisher {
    static func just(_ value: Output) -> AnyPublisher<Output, Failure> {
        Just(value)
            .setFailureType(to: Failure.self)
            .eraseToAnyPublisher()
    }
    static func empty() -> AnyPublisher<Output, Failure> {
        Empty()
            .setFailureType(to: Failure.self)
            .eraseToAnyPublisher()
    }
    
    static func fail(_ error: Failure) -> AnyPublisher<Output, Failure> {
        Fail(error: error).eraseToAnyPublisher()
    }
}
```

this could allow something like
```swift
func somePublisherValue() -> AnyPublisher<String, Never> {
.just("hey")
// instead of Just("hey").eraseToAnyPublisher()
}
```

This should also be super-useful in complex flatMaps with AnyPublishers as return values and some guard/let that always returns a Just/Fail

Is it something that could be useful or it's pointless? Am I missing something? Is it already there and I can't see it? :D 

Cheers and thanks (as usual) for the awesome work.

stefanomondino:
fixes #155 

I'm not sure if creating a `AnyPublisher.swift` file is the proper way to go, let me know :)

### DIFF
--- a/Sources/Convenience/AnyPublisher.swift
+++ b/Sources/Convenience/AnyPublisher.swift
@@ -1,0 +1,33 @@
+//
+//  AnyPublisher.swift
+//  CombineExt
+//
+//  Created by Stefano Mondino on 14/04/23.
+//  Copyright © 2020 Combine Community. All rights reserved.
+//
+
+#if canImport(Combine)
+import Combine
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+public extension AnyPublisher {
+    /// A wrapper around a type erased Just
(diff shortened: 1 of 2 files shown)


freak4pc:

PR #158, opened 2023-05-18
Feat: add an extension to subscribe a KeyValuePath to a Publisher

Oni-zerone:
Maybe someone could find this interesting? 
Just a way to subscribe a KeyValue path to a publisher in one line, having the possibility also to specify the Scheduler that you want to use.

Just an use example:
```swift
    let titleRelay = CurrentValueRelay("Await")
    awaitLabel
        .subscribe(\.text, to: titleRelay)
        .store(in: &cancellables)
```


freak4pc:

PR #162, opened 2023-07-01
Add codable conformance to Event

JackYoustra:
Only when `Output` and `Failure` are both also `Codable`


freak4pc:

Issue #167, opened 2024-02-10
Memory leak issue with PassthroughRelay


just1103:
Hello everyone,

I've been using CombineExt extensively in my projects and found it incredibly valuable.
Recently, I encountered a memory leak issue with `PassthroughRelay`. Upon replacing it with `PassthroughSubject`, the retain cycle disappeared.

It appears that `PassthroughRelay` fails to release its subscriptions upon receiving completion, as similarly described in this [PR #85](https://github.com/CombineCommunity/CombineExt/issues/85).

I've come up with a potential solution for this issue and confirmed its resolution in my project, but I'm not fully proficient with Combine. So, any comments or suggestions would be greatly appreciated. 

--- 

Original `Subscription`’s `cancel` method:
```swift
func cancel() {
    sink = nil
}
```

Solution:
```swift
// Call forceFinish() in Subscription’s cancel method. 
// And add self.sink = nil in forceFinish method.

func forceFinish() {
    self.sink?.shouldForwardCompletion = true
    self.sink?.receive(completion: .finished)
    self.sink = nil // added
}

func cancel() {
    forceFinish() // replaced
}
```

freak4pc:
Fixed on main, thanks!

PR #168, opened 2024-02-10
Fix memory leak issue with PassthroughRelay

just1103:
Fixed a memory leak with `PassthroughRelay` as described in this [issue #167](https://github.com/CombineCommunity/CombineExt/issues/167).

It appears that `PassthroughRelay` fails to release its subscriptions upon receiving completion, as similarly described in this [PR #85](https://github.com/CombineCommunity/CombineExt/issues/85).

I've confirmed the disappearance of the retain cycle with this solution in my project, but I'm not fully proficient with Combine. So, Any comments or suggestions would be greatly appreciated.

### DIFF
--- a/Sources/Relays/PassthroughRelay.swift
+++ b/Sources/Relays/PassthroughRelay.swift
@@ -78,8 +78,9 @@ private extension PassthroughRelay {
         }
 
         func forceFinish() {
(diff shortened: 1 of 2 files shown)


just1103:
Hello @freak4pc 
Could you please review this PR?

